PR #55, opened 2018-01-24
chore: Make all unit tests run in the browser

wmhilton:
If we want to get these tests running in the browser, we need to stop requiring all the pieces individually from `./dist/for-node/commands` and `./dist/for-node/utils` etc. This consolidates the import locations into just two. Next we'll have to build a browser bundle for `internal-apis.js`

Now it build a browser bundle for `internal-apis` called `internal.umd.min.js`. Also, implemented snapshot testing for the browser and a sort-of-isomorphic `makeFixture` utility. And switched from tape to Jasmine.


wmhilton:
grr CI

PR #92, opened 2018-03-14
Fix TypeScript 2.7.2 reported errors in the type declarations

TomasHubelbauer:
There were three types of problems:

`TS1040: 'async' modifier cannot be used in an ambient context`

This means the modifier `async` was included in the types incorrectly. If this was an interface and not a type declaration, you'd leave it up to the implementer to either use an `async` function or one that returns a `Promise` directly. In the same way, as long as the function returns a `Promise`, it's `await`able and it does not matter what the actual implementation is from the point of view of the type declaration.

`TS6137: Cannot import type declaration files. Consider importing 'events' instead of '@types/events'.`

This is a simple problem, we import `events` as we would, the fact that we import it only for types does not matter. TypeScript understands that.

`TS1254: A 'const' initializer in an ambient context must be a string or numeric literal.`

The problem here was that `=` was used instead of `:`, as if we were assigning a value and not declaring a type.

I think this shouldn't break anything in terms of older TypeScript versions etc. I am not sure anyone used the typings yet, so maybe I'm the first one to discover the issue.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,4 +1,4 @@
-import { EventEmitter } from "@types/events";
+import { EventEmitter } from "events";
 
 // Type definitions for isomorphic-git 0.x.x
 // Project: isomorphic-git
@@ -81,14 +81,14 @@ export interface MergeReport {
 
 /*~ If this module has methods, declare them as functions like so.
  */
-export async function add(args: {
+export function add(args: {
   fs: any,
   dir: string,
   gitdir?: string,
   filepath: string
 }): Promise<void>;
 
-export async function checkout(args: {
+export function checkout(args: {
   fs: any,
   dir: string,
   gitdir?: string,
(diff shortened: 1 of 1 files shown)


TomasHubelbauer:
I think if you run `tsc index.d.ts` to the CI, it should work as a test for the validity of the types.

wmhilton:
Thanks!!! This looks great! I don't actually use TypeScript, so I have been "testing" the typings in the vague sense that VS Code's autocomplete seems to pick up on them. I like the idea of testing the typings. Could we do it with something like "npx tsc src/index.d.ts"? I've got a bunch of test scripts in package-scripts.js where I've been managing them, we could add it there and make sure it runs in CI.

TomasHubelbauer:
Yeah for sure, `npm tsc src/index.d.ts` should work just fine. `0` if A-OK otherwise non-zero and it will pick up the version of TypeScript you decide to support by placing it to `devDependencies`, so the added benefit will be the existence of "documentation" of what the supported version is.

wmhilton:
Hmm. For some reason, I can't add commits to this PR. I've got a change that fixes the linting step and adds TypeScript as a devdependency, but I'll add that as a separate PR I guess.

PR #104, opened 2018-03-19
faster checkout

wmhilton:
Also fixes some issues with GitIndex where stat values exceeded what can fit in a uint32.


wmhilton:
There's something funny about these tests. It's creating some kind of error for iOS anyway.

wmhilton:
I'll have to try again later. I think I've made too many changes to attempt to rebase it at this point.

PR #105, opened 2018-03-19
fix: Switch from The Unlicense to The MIT License

wmhilton:
The consensus seems to be that using the standard MIT License will be more friendly to users and businesses than the Unlicense. I'm pretty sure that I, being the author, can relicense it this way. In jurisdictions where the public domain dedication was never valid, I still have the copyright. And in jurisdictions where the code became part of the public domain, I am free to include the text in my own work.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,10 +1,7 @@
-This is free and unencumbered software released into the public domain.
+Copyright 2017-2018 the 'isomorphic-git' authors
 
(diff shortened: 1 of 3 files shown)


joepie91:
I'm probably a bit late to the party, but the usual way to effectively dedicate things to the public domain when people have concerns about the license, is to use the [CC0](https://creativecommons.org/publicdomain/zero/1.0/). The CC licenses are widely accepted as valid licenses, including by businesses.

(For a bit more context: I license my own stuff under the WTFPL, and while I have no reason to believe that it's in any way legally invalid - even after speaking to lawyers about it - dual-licensing it with the CC0 stops people from complaining about the license, in my experience.)

PR #128, opened 2018-04-06
Emit status messages from git push as well as git fetch

wmhilton:
~~Blocker: I need to write some tests using the new karma-git-http-backend-middleware.~~
Done


wmhilton:
Nice, so far the blocking fs-extra.copy command is still fast enough that no connections are dropping. Hopefully it'll stay that way as more tests and larger repo fixtures are added.

PR #137, opened 2018-04-15
OpenPGPjs plugin (WIP)

wmhilton:
# See #218 for `signingHelper` architecture

If we do go with a plugin architecture, does it make sense to turn this into a monorepo or maintain them separately?


TomasHubelbauer:
Blessed plugins in tree and BYOP if preferred would be my vote. :-)

wmhilton:
I've been thinking about it, and while keeping "blessed" plugins in the same repo would be less work, putting them in their own repos would make them very useful as examples for others to follow. If we said, "making plugins is easy, look in the ./plugins folder" that might be less effective than saying "take a look at any of the plugins - see how the repo consists of package.json, index.js, and a README? it really is that simple".

If other people are going to write plugins (and I want to encourage that) they'll have to publish them as separate packages. So maybe we should do that too, just to make sure we're dogfooding it and sharing/empathizing with the plugin development experience.

TomasHubelbauer:
Yeah I didn't really consider that, I only thought about the maintenance part of the equation. I think it makes sense and agree.

wmhilton:
I'm thinking of keeping the plugin stuff on a separate 1.x branch for a bit. There are a couple small changes I want to make to the API that would be nice to lump together in a clean 1.x break. Notably, some of the functions return `Promise<String>` instead of `Promise<Object>` and so they can't be easily extended to return more information. For example, `git.commit()` returns just the SHA string, but now I wish it could optionally return the cryptographic payload (the text that gets signed) or the normalized, parsed commit object.

wmhilton:
Hmm. One thing I'm (re)discovering is that setting up testing (in node and the browser) for the plugin is annoying difficult. I might have to extract the test framework into its own package so that others can easily test their plugins.

wmhilton:
I just noticed the original reason that `sign` was its own function rather than an argument to `commit` was because of weight:

```js
-/**
- * Create a signed commit
- *
- * OpenPGP.js is a huge library and if you don't need to create or verify signed commits
- * you shouldn't be forced to include that weighty feature in your bundle. That's why this
- * is its own function.
```

Now that OpenPGP.js will be in its own plugin, we can simply make `signature: string` an argument to `commit`. That will spare us the extra file system writes, reparsing the commit, updating the `HEAD` reference twice, etc.

isysd:
> Now that OpenPGP.js will be in its own plugin, we can simply make signature: string an argument to commit. That will spare us the extra file system writes, reparsing the commit, updating the HEAD reference twice, etc.

I'm not sure this is true. The signature is over the commit itself, not the contents, so how can the signature string exist before the commit?

wmhilton:
Touche. I meant `signingKey: string`. Or possibly even just `sign: true` if we assume the plugin is bringing with it a keyring of sorts. Maybe `signingKey: id`? Then the private key never actually passes through the library, just the key id, opening up the potential for... IDK password manager / hardware supported / other better signing strategies.

wmhilton:
I shall use the same names as used in `.gitconfig`. Therefore the new arguments to `commit` will be: [`gpgSign:`](https://git-scm.com/docs/git-config#git-config-commitgpgSign)`boolean` and [`signingKey:`](https://git-scm.com/docs/git-config#git-config-usersigningKey)`string`.

PR #138, opened 2018-04-16
Start adding the server's side of the Git Smart HTTP protocol

wmhilton:


wmhilton:
I've got enough written that you can make a small HTTP server that handles clone requests!
Now to get it working over WebRTC ⚡️ and WebSockets 🔌 !

wmhilton:
Note: packObjects is returning incomplete packfiles because of #253. Need to fix that.

wmhilton:
Note to self: #253 has been merged so sendUploadPackResult might send complete packfiles now. Need to rebase and retest.

PR #142, opened 2018-04-23
New build toolchain using Webpack 4

wmhilton:
Webpack 4 actually worked surprisingly well out of the box, with only one tweak needed to fix `source-stream`. This PR looks at replacing the complex rollup+browserify building process with a simpler Webpack based one.

One reason for doing this is simply Webpack is extremely popular, and therefore supporting Webpack is a priority. Another reason is that PR #141 which updated rollup to the latest version broke the build. (Apparently rollup was not exporting CJS bundles... there's still raw `export` statements in them.) That's the second time that upgrading rollup broke the build, and I'm not a huge fan of tools that can't be updated without fear of things breaking. Hopefully Webpack will be more reliable in this manner. Lastly, complexity is a burden, and if the webpack process is simpler then I can get rid of a bunch of messy code.


wmhilton:
I ended up keeping rollup for bundling the ES Module and Node versions, and just using Webpack for the UMD build.

PR #149, opened 2018-04-29
Remove outdated esdoc and flow type annotations

wmhilton:


wmhilton:
Was bundled with #148

PR #156, opened 2018-05-02
is-buffer working in node 10.0.0 fixes #155

### LINKED ISSUE #155: checkout commit fails
Passing a commit hash instead of a branch to `checkout` results in an unexpected error. At least it is inconsistent with the standard git checkout command, though not the documentation for isomorphic-git.

Assume `83fa48b946132214648f0283a1e81fd4049fbb42` is a valid commit in the repository at `/keys/pgp`.

```
git.checkout({
   fs: fs, 
   dir: '/keys/pgp',
   ref: '83fa48b946132214648f0283a1e81fd4049fbb42'
})
```

Throws error:

```
 Error: invalid type passed to GitIndex constructor
    at new GitIndex (node_modules/isomorphic-git/dist/for-node/isomorphic-git/index.js:845:13)
    at Function.from (node_modules/isomorphic-git/dist/for-node/isomorphic-git/index.js:849:12)
    at node_modules/isomorphic-git/dist/for-node/isomorphic-git/index.js:1827:26
    at process._tickCallback (internal/process/next_tick.js:178:7)
```

Expected behavior would be to set HEAD to 83fa48b946132214648f0283a1e81fd4049fbb42.

Engine: nodeJS v10.0.0

Version: 0.11.2

isysd:
There appears to be a superficial difference between two types of buffers being used which caused `isBuffer()` to give false negatives, and also for `parseBuffer` to fail the checksum test.

Honestly I don't fully understand the problem, but it used to fail in node v10.0.0 as per #155 but this patch makes it work.

### DIFF
--- a/src/models/GitIndex.js
+++ b/src/models/GitIndex.js
@@ -118,7 +118,7 @@ export class GitIndex {
   constructor (index /*: any */) {
     this._dirty = false
     if (isBuffer(index)) {
-      this._entries = parseBuffer(buffer.Buffer(index))
+      this._entries = parseBuffer(Buffer(index))
     } else if (index === null) {
       this._entries = new Map()
     } else {
--- a/src/models/GitIndex.js
+++ b/src/models/GitIndex.js
@@ -3,6 +3,7 @@ import { Buffer } from 'buffer'
 import BufferCursor from 'buffercursor'
 import sortby from 'lodash.sortby'
 import shasum from 'shasum'
+import isBuffer from 'is-buffer'
 
 const MAX_UINT32 = 2 ** 32
 /*::
@@ -116,8 +117,8 @@ export class GitIndex {
    */
   constructor (index /*: any */) {
     this._dirty = false
-    if (Buffer.isBuffer(index)) {
-      this._entries = parseBuffer(index)
+    if (isBuffer(index)) {
+      this._entries = parseBuffer(buffer.Buffer(index))
     } else if (index === null) {
       this._entries = new Map()
     } else {
--- a/src/models/GitTree.js
+++ b/src/models/GitTree.js
@@ -1,5 +1,6 @@
 // @flow
 import { Buffer } from 'buffer'
+import isBuffer from 'is-buffer'
 
 /*::
 type TreeEntry = {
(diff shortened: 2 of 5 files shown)


src/models/GitIndex.js
```diff
@@ -116,8 +117,8 @@ export class GitIndex {
    */
   constructor (index /*: any */) {
     this._dirty = false
-    if (Buffer.isBuffer(index)) {
-      this._entries = parseBuffer(index)
+    if (isBuffer(index)) {
+      this._entries = parseBuffer(Buffer(index))
```
wmhilton: This can be just `this._entries = parseBuffer(index)`. We just confirmed it is a buffer, no need to re-create it. If we did need to create a buffer from say a string, the prefered way not is `Buffer.from(index)` since `Buffer(index)` got deprecated and the standard linter won't allow it.

src/models/GitTree.js
```diff
@@ -67,8 +68,8 @@ export class GitTree {
   _entries: Array<TreeEntry>
   */
   constructor (entries /*: any */) {
-    if (Buffer.isBuffer(entries)) {
-      this._entries = parseBuffer(entries)
+    if (isBuffer(entries)) {
+      this._entries = parseBuffer(Buffer(entries))
```
wmhilton: Same here, `this._entries = parseBuffer(entries)` ought to work, and won't cause the linter to throw an error.

src/models/GitIndex.js
isysd: If you do this then it fails the checksum. Not sure why. :/

wmhilton: Hmm, that's suspect. I'm thinking there may be an easier way out of this mess. With rollup I needed to specify a buffer module for the browser, but now that the build system uses Webpack it may not be needed anymore. I'm tempted to try just removing all the `import { Buffer } from 'buffer'`lines and see if that fixes this problem.

isysd: Definitely give it a shot! It is clearly a superficial difference between Buffer packages, and any way that results in a single package being used would surely resolve it.

wmhilton:
Closing in favor of #162 which also solved the problem and simplified things

PR #162, opened 2018-05-07
feat: Remove the direct dependency on the Buffer module

wmhilton:


isysd:
Awesome, thanks @wmhilton! I'll test this out and confirm it fixes the issue in #155. :+1:

PR #170, opened 2018-05-10
chore: cache node_modules

DanielRuf:

PR #172, opened 2018-05-10
chore: use npm ci

DanielRuf:
`npm ci` is much faster and is meant for CI. This PR is meant for evaluating the new ci command and testig the performance and compare the logs of Travis CI.

https://docs.npmjs.com/cli/ci
http://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,10 @@ cache:
     - ~/.npm
 notifications:
   email: false
+before_install:
+  - npm i -g npm@latest
+install:
+  - npm ci
 node_js:
   - '10'
 before_script:


wmhilton:
Of your three perf-related PRs this one looks like it could have the biggest impact! (installs in 48s instead of 60s)! However running "npm i -g npm@latest" first adds 10 seconds, so we're back where we started (58s).

DanielRuf:
> Of your three perf-related PRs this one looks like it could have the biggest impact! (installs in 48s instead of 60s)! However running "npm i -g npm@latest" first adds 10 seconds, so we're back where we started (58s).

Exactly. We could try to add the npm directory to the cache. Not sure if this would help much.

What do you think?

DanielRuf:
Closing in favor of https://github.com/isomorphic-git/isomorphic-git/pull/170.

We could also cache `$HOME/.npm` but still `npm ci` deleted `node_modules` and reinstalls the dependencies.

PR #173, opened 2018-05-10
chore(pkg): Add sideEffects: false flag

TheLarkInn:
Based on the reviewed module graph, it appears that there are no side effects created against each other. (For more information on what I mean here: https://stackoverflow.com/a/49203452/1624684). 

This change will enable webpack4 to perform even deeper treeshaking and dead code elimination out of the box. (Without any breaking change).

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
   "bin": {
     "isogit": "./cli.js"
   },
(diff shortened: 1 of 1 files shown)


wmhilton:
Awesome sauce!

Issue #108, opened 2018-03-20
Error when pushing without "ref"


juancampa:
The documentation says that `push` doesn't require a `ref` but I get the following if I don't specify it:

```
Cannot read property 'startsWith' of undefined
    at Object.push (commands.js:841)
```

The error goes away if I pass a `ref`.

Thanks for this amazing library

juancampa:
Documentation says

> Which branch to push. By default this is the currently checked out branch.

TomasHubelbauer:
I would like to take a look at this this weekend, so I am assigning myself (hopefully that's okay?). If you get to it sooner or just want to do it yourself, please feel free to unassign me @wmhilton.

wmhilton:
That's amazingly OK to assign yourself! To figure out what the current branch is you can use GitRefManager.resolve on 'HEAD' with a depth limit of... 2, I think? I should probably add a helper function to GitRefManager for getting the current branch.

jcubic:
I use code like this to get the branch:

```javascript
    async function gitBranch(filepath) {
        try {
            var dir = await git.findRoot({fs, filepath});
            var ref = await git.resolveRef({fs, dir, ref: 'HEAD', depth: 1});
            return ref.match(/ref: refs\/heads\/(.*)/)[1];
        } catch(e) {
        }
    }
```

TomasHubelbauer:
Actually @jcubic I am unable to work on this at the moment due to time reasons, would you like to take a shot yourself?

PR #216, opened 2018-05-20
push without ref #108

jcubic:
Adding git.push without ref + tests - fixing issue #108


wmhilton:
Nice!! Sorry I didn't see this sooner 😀 This is great!

PR #217, opened 2018-05-20
(WIP) Eslint error lines (fixes #212)

### LINKED ISSUE #212: Show API function name in error message
So now I have complicated app with lot of calls for git functions and I use one catch at the end but I don't know what caused the issue it would be usefull if each error message show exact api method that caused the error so it's easy to find the problem.

This may be problematic because they are other functions, that not exposed, that can cause errors.

wmhilton:
This is a proof-of-concept test run of my weekend invention, https://github.com/wmhilton/eslint-plugin-awesome-errors, that would make it possible to automatically include the file name, line number, and an Error Code in all error messages.

Note: Theoretically I could extend it to print the name of the current function. Which would be useful as a "callstack alternative" because ~~it would just show the function a user _actually_ called instead of everything in the call stack.~~ Edit: Oooh... that wouldn't always work, because some exported functions call each other (clone calls fetch, etc) and so errors in `fetch` can't be hardcoded to say "fetch" because that would confuse the user if the user called `clone`. I think the solution to this will be to refactor the functions to take a hidden `caller` argument that can be displayed in error messages.

## Philosophy
Premise: There are 3 basic approaches to solving errors:
- a) figure out what your application (the code calling isomorphic-git) is doing wrong ("user error")
- b) figure out what the library (the code isomorphic-git runs) is doing wrong ("a bug")
- c) give up and google it

Exceptions that are thrown by the library (not handled internally) should be designed to help the reader employee these 3 strategies.

## Proposals
To this end, I propose the following for designing Awesome Errors (name TBD):
- **application debugging**: 
  - Perform assertions on all function arguments and throw errors explaining the correct usage. E.g. `Error("fs is a required argument")`.
  - Do the refactoring work described above so that errors that occur in say, GitObjectManager, will display the name of the function the application code actually called.
- **library debugging**:
  - Include the filename and line number where the error originated. Rather than relying on call stacks, I'm suggesting that errors directly embed the filename and line number into the error message string, using a (super cool) eslint plugin that will automatically fix the line numbers whenever errors move around.
- **searching for solutions**:
  - Assign each error a permanent code so users can more easily search to find other users encountering the same error.
    - Note: While assigning meaningful codes (like HTTP statuses) that can be reused throughout an application was my initial design goal, I'm now leaning away from that idea. Why? Because the root cause of a "file not found" error during `fetch` is probably completely different from the root cause of a "file not found" error during `checkout`. So while it seems appealing to use the same code (e.g. "E100 FileNotFound") wherever a "file not found" error is needed, it actually makes searching for a solution harder because you have to wade through the irrelevant search results for "isomorphic-git E100 FileNotFound" errors to find the relevant ones. (This is hypothetical - isomorphic-git is too young to have tons of stackoverflow and github bug reports - but is based on my own experience as a developer/user.)
  - These codes could (and should) also be documented on the website.


jcubic:
What about stack trace but only for API functions? Are there other functions that call each other that will created stack with more then 2 items?

wmhilton:
... Yeah, yes I guess there are. Typically a command will call into a manager, which may call into another manager, which calls into a model

wmhilton:
Thinking about the [realistic user scenario](https://github.com/isomorphic-git/isomorphic-git/issues/226) here, I'm wondering about how to delimit error codes. For instance:
```js
error.message.includes('E34')
```
would erroneously match `new Error('E340')` and `new Error('ENGINE34')`. But using spaces, e.g.
```js
error.message.include(' E34 ')
```
seems hacky as well because it would break if say, there was no leading or trailing space.

Should we include errors in brackets (in a form with a space so that that would be not get confused with array indices) like:
```js
new Error('[Error 34]')
```

Of course, that's not foolproof because users might just ignore the brackets and use `error.message.include('Error 34')` anyway. What about:

```js
new Error('Error[34]')
```

Or we could go with a different prefix-suffix system altogether, like:
```js
new Error('#34E')
```

So far I think I like the `'Error[34]'` style the best... what do you think?

wmhilton:
Should I make the matter moot by providing an error ~~parser function~~ comparison utility to accompany `isomorphic-git`? E.g.

```js
} catch (err) {
  if (git.matchError(err, git.ERROR_FAILED_OID_READ)) {
```

wmhilton:
or how about,
```js
} catch (err) {
  if (git.catch(err) === git.ERROR_FAILED_OID_READ) {
```

and yes I'm just riffing at this point. Flow-of-consciousness style.

jcubic:
Right now what I would have is just `err.caller` that show API function. It can even be `git.catch(err).caller` it can show first API function that throw error so it will definitely be the one that user called. The exact line that throw error may be useful when users will have issue and will ask question on SO or in github issues.

PR #257, opened 2018-06-02
Update to latest stream-http to remove user-agent restriction

wmhilton:


jcubic:
Did you modify package.json? because only package-lock.json is showing in changes

PR #266, opened 2018-06-07
fix: only update remote refs that we actually have downloaded

wmhilton:
## I'm fixing a bug

- [x] make a fix commit "fix: [Description of fix]"

This is a (surprisingly complicated) set of changes so that when you do `isogit clone --singleBranch` you end up with the same entries in `remotes/origin` as you do when you run `git clone --single-branch`. 
- It doesn't update the references until _after_ the packfile is downloaded. Before, it erroneously reported having all the refs!
- It now saves the remote `HEAD` to `remotes/origin/HEAD` if you clone the default branch, even though that's not in the git config refspec (🤷‍♂️ ).
- If you clone with `singleBranch: true` it only saves the refs (and symrefs) for the branch you download.

**This should fix some of the spurious errors with `clone` and `checkout` we have been seeing!**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ cache:
 notifications:
   email: false
 node_js:
-  - '10'
+  - '10.3.0'
 install:
   - npm i
 before_script:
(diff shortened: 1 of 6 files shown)


wmhilton:
This "Jest hanging right at the end of the tests" is driving me nuts.

PR #272, opened 2018-06-09
(WIP) Refactor with Functions

wmhilton:
I'm hoping that switching from classes with static methods to just plain old function named exports, I'll be able to achieve tighter tree-shaking. In particular, I want to be able to sever read functionality from write functionality, and currently those are bound together as methods on the same class, e.g. `GitBlaw.read()` and `GitBlaw.write()`.

Also, this refactoring should simplify the plugin system design, as I will ultimately need to centralize all state artifacts in one place. Currently state is spread out among files, making it harder to tree-shake, I think.


wmhilton:
I'm going to redo this more methodically. Some of the changes I thought were inconsequential turned out to negatively impact bundlesize. I'll break this task down into smaller tasks in separate PRs.

PR #273, opened 2018-06-09
feat: Simplify API for authenticating with personal access tokens

### LINKED ISSUE #242: Auth with personal access token
Is it possible to fetch with a Github Personal Access Token? I don't want to have to create a new account for the server that would be using this module, or add a credential with more permission than it needs. Seems fairly straightforward based on [this](https://blog.github.com/2012-09-21-easier-builds-and-deployments-using-git-over-https-and-oauth/) but would like some insight before going down the rabbit hole.

screen shot 2018-05-27 at 8 30 14 pm

wmhilton:
This aims to resolve #242 in the most satisfactory way. I won't merge this until some people have had time to look at the [truth table](https://github.com/isomorphic-git/isomorphic-git/issues/242#issuecomment-395990079) and decide whether the argument precedence makes sense. It needs to handle subtle cases like:
- If both `token` and `password` are specificed, should it throw an Error?
- If `oauth2format` is specified, do we simply ignore `username` and `password`? Or do we throw an error?
- If `token` is specified without a username, do we use Github's shorthand? Or should we assume that's an error and require `username` be `''`?


src/index.d.ts
```diff
@@ -279,8 +287,10 @@ export function push(args: {
   ref?: string,
   remote?: string,
   url?: string,
-  authUsername?: string,
-  authPassword?: string
+  username?: string,
+  password?: string,
+  token?: string,
+  oauth2format?: string,
```
TomasHubelbauer: Should we specifically type `oauth2format` as `'github'|'bitbucket'|'gitlab'`? It will error in compile time, which will prevent from deferring the bug discovery til runtime, but we will still have the `default` check for non-TS use.

src/utils/calculateBasicAuthUsernamePasswordPair.js
```diff
@@ -0,0 +1,32 @@
...
+    .map(Number)
+    .join('')
+  // See helpful truth table in documentation
+  // prettier-ignore
+  switch (key) {
+    case '0000': return null
+    case '1000': throw new Error(`Missing password or token`)
+    case '0100': throw new Error(`Missing username`)
+    case '1100': return { username, password }
+    case '0010': return { username: token, password: '' }
```
TomasHubelbauer: This deserves at least a comment. :-)

TomasHubelbauer: That said I am not opposed to this shortcut, although I do find myself wishing we could only limit it to GitHub smarter - but I don't see how we could do it. I figured host check in the URL would work, but that would fail with enterprise GitHub deployments. So I think let's leave this in and see if people get confused by this or not.

src/utils/calculateBasicAuthUsernamePasswordPair.js
```diff
@@ -0,0 +1,32 @@
...
+    case '1010': return { username, password: token }
+    case '0110': throw new Error(`Missing username`)
```
TomasHubelbauer: Is this a good error? If I go fix this and add a username, I will end up with a new error - `1110`.

src/utils/calculateBasicAuthUsernamePasswordPair.js
```diff
@@ -0,0 +1,32 @@
...
+    case '1110': throw new Error(`Can't have both password and token`)
+    case '0001': throw new Error(`Missing token`)
+    case '1001': throw new Error(`Missing token`)
+    case '0101': throw new Error(`Missing token`)
+    case '1101': throw new Error(`Missing token`)
+    case '0011': return oauth2(oauth2format, token)
```
TomasHubelbauer: I think the only valid `??11` state should be `0011`, others ignore `username` and/or `password` even though they are in and that to me indicates the user either made a mistake or expects them to influence the Oath2 flow whereas only token ends up being used and it's not obvious that we'd ignore the user name and password silently when token is passed.

src/utils/calculateBasicAuthUsernamePasswordPair.js
```diff
@@ -0,0 +1,32 @@
...
+  token,
+  oauth2format
+}) {
+  const key = [!!username, !!password, !!token, !!oauth2format]
+    .map(Number)
+    .join('')
+  // See helpful truth table in documentation
+  // prettier-ignore
+  switch (key) {
+    case '0000': return null
```
jcubic: This switch with flags is very nice solution.

src/utils/calculateBasicAuthUsernamePasswordPair.js
wmhilton: I know right?! That drove me nuts, but I couldn't think of a good alternative at the time. If I made it the same error as 1110, then when they remove either password or token... oh hang on. Removing password would actually be valid because of the Github shorthand.

src/utils/calculateBasicAuthUsernamePasswordPair.js
wmhilton: Yeah... I guess for some reason I thought users might want to support both in the same call, and if an OAuth2 token is available let that take precedence. But now that I'm thinking about it... they'd probably be very different code paths, because obtaining an OAuth2 is no trivial task. So that scenario where it's convenient probably doesn't exist. Probably wiser to throw errors rather than silently ignore.

Actually, that might just be a good blanket policy: throw an error when combinations of arguments don't make sense, rather than silently ignore arguments. Might be annoying to users, but it also guarantees they are aware of the valid combination.

wmhilton:
Thank you @TomasHubelbaur and @jcubic for taking the time to review this!!!

wmhilton:
@TomasHubelbauer I'm pretty confident that you'll find the changes I've made as a result of your suggestions satisfying, so I'm going to go ahead and merge this.

src/index.d.ts
TomasHubelbauer: Yeah, explicit deprecation is the way to go I think. :-)

PR #274, opened 2018-06-10
fix: Fixed bug where merge commits appeared to only have one parent (fixes #253)

### LINKED ISSUE #253: git.log needs to trace all merge parents
Right now, git.log is only following the first parent. This causes a bug in building packfiles for sendUploadPackResult because it is not including the commits from the merged branches.

wmhilton:


wmhilton:
OK so for the record, for some reason this change:
image

is the difference between a 3 kilobyte jump and a 29 byte jump.

wmhilton:
I think it's to do with circular dependencies.

PR #286, opened 2018-06-17
see if using `export * from` changes bundle size

wmhilton:
Null Hypothesis: does not affect bundle size
Alternate hypothesis: size will go up
Bizarre hypothesis: size will go down


wmhilton:
ok... it's 7 bytes larger? "export " is 7 bytes... coincidence?

wmhilton:
ooh, i need to make sure it's not exporting something that I only want exported to `internal-apis.js`

wmhilton:
Yup, yup, `push.js` is exporting more than just `push`. I need to refactor that first.

wmhilton:
OK... so now it's 110B smaller. I guess the bizarre hypothesis wins.

PR #287, opened 2018-06-18
(WIP) formalize Error Codes

wmhilton:
a variation on #217 with a more traditional approach

* * *

This adds a new model, GitError. A GitError extends the base Error class in JavaScript, with some new properties and changes some existing properties:
- a `caller` string (see #262)
- a `code` string
- a `data` object
- a `message` string; however that string is now declared in a central location, templated with `{ placeholders }` filled in by `data`, and translated based on locale. (Hopefully volunteers will provide translations in other languages!)
- a `toJSON` method; returns { code, data, caller, message }
- a `toString` method that shows the message and stack trace; however it now prints the error code instead of just "Error:" at the beginning.

# Example
```js
import { E, GitError } from '../models';
try {
  throw new GitError(E.FileReadError, { filepath: 'foobar.txt' })
} catch (err) {
  if (err.code === E.FileReadError) {
    console.log(err.toString())    // FileReadError: Could not read file "foobar.txt"
    console.log(err.data.filepath) // foobar.txt
    console.log(err.toJSON())      /* {
                                      caller: undefined,
                                      code: "FileReadError",
                                      data: { filepath: "foobar.txt" },
                                      message: 'Could not read file "foobar.txt"' 
                                    } */
  } else {
    console.log('unhandled error', err)
}
```


wmhilton:
There are many `throw new Error` in my code. I'm like... only halfway done 😩

erdahuja:
@wmhilton what help do you need in this?
Also, i was thinking if we add 
```
  static of(code: Codes) {
    throw new Error(code);
  }
```
in Error class, it will save lot of typing of 'throw new Error()'
 instead it will just be Error.of(code).

wmhilton:
@erdahuja I thought about something like that, but you need to call `throw` where you actually have the error to get a useful stack trace. Otherwise, it just appears like all your errors come from the `of` function.

Typing `throw new GitError` isn't a problem. It's just going through them all and moving the error messages to GitError.js so we can provide translations and fixed error codes.

erdahuja:
Okay @wmhilton , How can i proceed with it to help you out? @wmhilton I couldn't find a good place to start contributing during inital commits, to get hang of code and feel confident about it?

Also I am unable to fetch this branch because of permission issues.

wmhilton:
This probably isn't a good one to start with, because we'd likely have merge conflicts if we both tried to write to this branch. Maybe this weekend I'll do a livecoding session and make an official getting started tutorial!

wmhilton:
@jcubic you probably have the most experience catching errors I throw ⚾️. What do you think of this?

src/models/GitIndex.js
```diff
@@ -55,18 +56,24 @@ function parseBuffer (buffer) {
   let shaComputed = shasum(buffer.slice(0, -20))
   let shaClaimed = buffer.slice(-20).toString('hex')
   if (shaClaimed !== shaComputed) {
-    throw new Error(
-      `Invalid checksum in GitIndex buffer: expected ${shaClaimed} but saw ${shaComputed}`
-    )
+    throw new GitError(E.InternalFail, {
+      message: `Invalid checksum in GitIndex buffer: expected ${shaClaimed} but saw ${shaComputed}`
```
jcubic: GitError constructor as I undestand execute messages function that return translation of the message. Isn't you end up with `error.message` and `error.data.message`?

src/models/GitError.js
```diff
@@ -0,0 +1,147 @@
...
+  MixUsernameOauth2formatTokenError: `MixUsernameOauth2formatTokenError`,
+  MixPasswordOauth2formatTokenError: `MixPasswordOauth2formatTokenError`,
+  MixUsernamePasswordOauth2formatTokenError: `MixUsernamePasswordOauth2formatTokenError`
+}
+
+export class GitError extends Error {
+  constructor (code, data) {
+    super()
+    this.name = code
+    this.code = code
```
jcubic: why two different fields with the same value?

src/models/GitError.js
```diff
@@ -0,0 +1,147 @@
...
+    this.data = data
+    this.message = messages[code](data || {})
```
jcubic: Maybe this should be 
```javascript
this.message = data.message || messages[code](data || {})
delete data.message; // but first data need to be cloned to don't delete user field
this.data = data
```

src/models/GitIndex.js
wmhilton: Eh... true, but I don't expect a lot of users to ever encounter these InternalFail errors.

src/models/GitError.js
wmhilton: Ah! This one is interesting. The default `Error` class apparently uses the `.name` value in it's `.toString()` implementation. So by setting `.name` it magically starts showing up as "MyError: blaw blaw" instead of "Error: blaw blaw" in the console.

But... I felt like relying on `.name` was a little risky (ES6 will "automagically" assign `.name` to functions and classes in many cases. Try it `var bob = () => ''; bob.name`) so I stuck with `.code` as the "official" property to look at.

src/models/GitError.js
wmhilton: Ooooh. Oooh. Is it _too_ clever? I like it.

wmhilton: I think it's _too_ clever. I think I'll just make a second constructor called `GitInternalError`.

wmhilton:
Yes, I'd like to further organize the errors into categories that help users and provide suggestions. I started by trying to separate Errors from Failures; conceptually an Error is something the user might be able to fix, but a Failure is something they probably can't. But it's hard to draw solid lines around these categories.

wmhilton:
> One thing I would expose GitError constructor to the users maybe git.Error.

Why though? If other code starts throwing errors that look like they are coming from isomorphic-git it would be confusing. I expose all the error _codes_ as

```import { E } from 'isomorphic-git'```

**Comment by jcubic** [comment]
> > One thing I would expose GitError constructor to the users maybe git.Error.
> 
> Why though? If other code starts throwing errors that look like they are coming from isomorphic-git it would be confusing. I expose all the error codes as
> 
> import { E } from 'isomorphic-git'

what about the case:
```javascript
try {
  //lot of code
  git.fn();
} catch(e) {
  if (e instanceof git.Error) {
     // isomorphic git errors
  } else {
    // other errors;
  }
}
```
maybe you can use some factory trick that will allow you to create instance of the error but don't allow to create instance directly, don't know if this is possible in JavaScript. In C++ you can have private constructors.

erdahuja:
We can do this to separate Errors and Failures-

```
const codesGroup = {
  Error: [
    'Error 1',
  ],
  Failure: [
    'Failure 2',
  ],
};

function getCodeGroup(code) {
  return Object.keys(codesGroup).find((key) => {
    const codesForKey = codesGroup[key];
    return codesForKey.indexOf(code) !== -1;
  }) || 'unknown';
}

const codeGroup = getCodeGroup(code);
```

wmhilton:
> what about the case:

I would recommend:

```js
try {
  //lot of code
  git.fn();
} catch(e) {
  if (typeof e.caller === 'string' && e.caller.startsWith('git')) {
     // isomorphic git errors
  } else {
    // other errors;
  }
}

src/models/GitError.js
wmhilton: On third thought... I don't see why I should worry about this at all. I think it's fine the way it is.

wmhilton:
> We can do this to separate Errors and Failures-

Let's not overcomplicate things. If it gets to be a mess and we need to subdivide them into groups we can do that later.

PR #295, opened 2018-06-24
perf: do less parsing of idx files to save time

wmhilton:
the branch name is a misnomer. True lazy parsing would only access the file on readObject and only parse the relevant entries in the idx file. This is just me speeding up the existing eager idx parser.


TomasHubelbauer:
Sidenote: are there any packages for testing performance for NodeJS? Something that would warn on an unusually large slowdown for some tests (say >5% worse across changes)?

wmhilton:
That would be a great idea.

PR #305, opened 2018-07-02
perf: speed up packfile parser and reduce bundle size

wmhilton:
Replaced git-list-pack with my own code using pako to find the object offsets in the packfile.


wmhilton:
Speeds up indexPack by at least a factor of 2.

PR #317, opened 2018-07-05
Fix the base branch

TomasHubelbauer:


TomasHubelbauer:
As far as I can see the CI problem seems to be with SauceLabs browsers disconnecting?

wmhilton:
They do that. I really need to perfect that script that's supposed to skip the tests for README changes.

PR #326, opened 2018-07-12
fix: cache stat values after calling git.status

wmhilton:
I think to see real perf improvements though, I need to support calling `git.status` on whole directories at a time (because status requires walking the HEAD commit, and it's better to walk just once).

### DIFF
--- a/src/commands/status.js
+++ b/src/commands/status.js
@@ -77,13 +77,18 @@ export async function status ({
         })
         // If the oid in the index === working dir oid but stats differed update cache
         if (I && indexEntry.oid === workdirOid) {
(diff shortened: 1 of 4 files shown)


wmhilton:
It should be noted that this may not be saving the index compatibly. I'm finding my git directory gets confused and has trouble making commits until I delete `.git/index` and create a fresh one.

wmhilton:
OK, I have no idea how I broke it. RangeError: "value" argument is out of bounds?

wmhilton:
Figured it out. BrowserFS returns -1 for `size` in fs.stat for the HTTP Request backed file system if no Content-Length header is presented.

PR #329, opened 2018-07-12
Update README.md

wmhilton:
Got an email from Sauce Labs complaining that our test results weren't public. Adding a Sauce Labs branded link.


wmhilton:
It's saying it can't find "nps". Which means something about "npm install" changed. Which is weird because I've got a `package-lock.json`. And it doesn't look like Travis is using a different version of node or npm than it was for tests that worked fine before.

PR #331, opened 2018-07-13
see if removing opencollective postinstall script fixes runkit

wmhilton:
I noticed as I was updating my [tutorial](https://runkit.com/wmhilton/5a89cd94303ded001277664c) that the ‘isomorphic-git’ package is out of date on runkit. Runkit is stuck at version v0.12.3.

I contacted Runkit and was told:
>  Unfortunately, we've temporarily disabled natively-compiled packages to prevent abuse and attacks on our infrastructure. We're eagerly looking forward to re-enabling this and re-processing native-code packages, but unfortunately do not have a timeline just yet.

Obviously, isomorphic-git is pure JavaScript. Jest has some native modules (for real, it's sad) however v0.12.3 uses Jest as well so that's probably not it. Two packages were added in the v0.13.0 release:
- git-clean-ref
- opencollective

OpenCollective *does* have a postinstall script that prints a message. So that might be what's triggering this ban? Anyway, it's been up for 2 months so I don't think we'll lose any valuable donations by not showing a banner ad on install. Kinda obnoxious really. So even if it doesn't fix the Runkit problem I might leave out the opencollective dependency.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1016,17 +1016,20 @@
     "ansi-escapes": {
       "version": "1.4.0",
       "resolved": "https://registry.npmjs.org/ansi-escapes/-/ansi-escapes-1.4.0.tgz",
-      "integrity": "sha1-06ioOzGapneTZisT52HHkRQiMG4="
+      "integrity": "sha1-06ioOzGapneTZisT52HHkRQiMG4=",
+      "dev": true
     },
     "ansi-regex": {
       "version": "2.1.1",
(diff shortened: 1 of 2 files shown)


wmhilton:
IT FIXED RUNKIT! Woohoo! The [quick start tutorial](https://runkit.com/wmhilton/isomorphic-git-quick-start) has been updated to version isomorphic-git@0.22.3.

PR #341, opened 2018-07-21
push --force

### LINKED ISSUE #219: git push force
So now from #129 I have function that restore git repo from `HEAD~1` but when I have pushed the commit I to update remote I need to use `git push -f`.

if this is how push works every time then I need to a way to check if push will overwrite commits so I would be able to reject it when user don't use force.

wmhilton:
This adds a new command, `git.isDescendent`, and a `force` option to `git.push`
Resolves #219


hsablonniere:
@wmhilton It works on our context!!

PR #349, opened 2018-07-26
Issue 345 list remotes

### LINKED ISSUE #345: New feature: list git remotes
Hey,

I would like a new command to list git remotes.
It seems like the whole list is already in the config file.

WDYT?

### LINKED ISSUE #359: Every time config is edited, extra newline is added to it
I'm using node 10 and every time I edit config with config() function, it adds empty line to the config for each edit. After a while, git config has hundreds of empty lines all over the place.

hsablonniere:
This is work in progress to fix #345 
also fixes #359 

## I'm adding a new command:

- [x] add TypeScript definition in `src/index.d.ts`
- [x] add command to README list of commands
- [x] ~~make a feature commit "feat: Added 'X' command"~~ _TODO: This should be removed from the PR template, because it's done during the Squash and Merge button phase._


hsablonniere:
* I'm not decided on the object vs array for the return.
* I still miss stuffs (docs, typing...)
* Maybe we need deeper tests
* Maybe we need to refactor the way we parse the file because the current state of this introduces lots of code duplication
* We also need to consider how we're going to update a add remote a remote.

wmhilton:
Definitely on the right track! Love that you even added a test.

Yes, I think returning the full `fetch` refspec is probably more than what users are looking for. I'm inclined to transform it into something more like `[{name, url}, ...]`.

To add a remote, it should be as simple as
```js
let name = 'foobar'
let url = 'https://github.com/foobar'
const config = await GitConfigManager.get({ fs, gitdir })
await config.set(`remote.${name}.url`, url)
await config.set(`remote.${name}.fetch`, `+refs/heads/*:refs/remotes/${name}/*`)
```

To delete a remote:
```js
let name = 'foobar'
const config = await GitConfigManager.get({ fs, gitdir })
await config.set(`remote.${name}.url`, undefined)
await config.set(`remote.${name}.fetch`, undefined)
```

Although... now that I'm thinking about it that won't totally work, because we need a way to delete the empty section header. But preferably only if the section is truly empty. Maybe for now don't worry about it. The `git remote` command ignores `[remote "foobar"]` if it is an empty section, so we can do the same with listRemotes.

wmhilton:
We should also throw in a `git-clean-ref` check to validate the remote `name` similar to how we do for branch [here](https://github.com/hsablonniere/isomorphic-git/blob/1de95bfe1207325aa5558189cfac5c6667030262/src/commands/branch.js#L28)

hsablonniere:
@wmhilton I pushed a new proposition to "parse" the config file ;-)

As you can see in the different commits:

1. I added a full test suite for the GitConfig model so I would be iso with my new impl
2. I refatored the code with a "immutable" style (which is a bit more verbose)
3. I have a commit with a more mutalbe impl (we can still remove this commit, up to you)
4. I took the opportunity to support comments

This refactor allowed me to code the `listRemotes easily` but it will have a bigger impact on how simple remote creation and deletion will work.

hsablonniere:
@wmhilton Why are all the methods in `GitConfig` async?

wmhilton:
> Why are all the methods in GitConfig async?

forward compatibility. Everything is async, so that it can be moved off the main thread into a WebWorker easily. Or read from disk on demand. Or whatever.

wmhilton:
I like it! We preserve the magical ability to edit just the lines that need to be modified, but also (sort of) parse it by smearing the section metadata across the lines in the constructor.

wmhilton:
I'll check it out thoroughly this evening. 👍

hsablonniere:
I just tried a few tests and also checked with the classic git CLI client.

* odd number of quotes in a value => error
* even number of quotes => quotes are just stripped
  * you can have parts in an out of quotes
  * you can have multiple quoted parts

wmhilton:
You can also escape quotes and comment markers using a backslash inside a quotation though.
```
foo = "this is \"tricky\"; and #not a comment"
```

hsablonniere:
@wmhilton I updated lots of stuffs sorry ;-)

The problem is, when you start to read the git config doc, you just want to write tests and features :smile:

hsablonniere:
I think we improvemed the % of support on the config a lot (even if that was not really the most important priority) but maybe I went too far (too much code etc...) Let me know what you think.

Here's a list of stuffs we don't support:

* odd number of quotes => error
  * I think we should do this one
* multiple line values
* includes...

wmhilton:
Ha! I think this is great! Remove the "WIP" label from the title and let me merge it already! ❤️

wmhilton:
I'm going to take care of update the `.all-contributorsrc` file here, because the CLI that automates adding users doesn't seem to be working. (I wonder if it's just me?) And it is a few contributors behind.

wmhilton:
OK, hang on, I retested and #359 is still an issue.

hsablonniere:
@wmhilton I removed the WIP, do you take care of the docs?

wmhilton:
Yes I'll be happy to update the docs. 😄

Issue #284, opened 2018-06-16
Can not know staged removing files without transaction


mizchi:
## tl;dr;

I can not know removing files in git without `fs` transaction clue.
isomorphic-git may need that state directly.

## Procedure

- git.init /repo
- fs.writeFile /repo/a
- git.add /repo/a
- git.commit(...) 
- fs.unlink /repo/a
- git.remove /repo/a
- git.status /repo/a => 'deleted'

In this case, if I know `/repo/a` path, I can get `/repo/a` is deleted.

- `git.listFiles` does not know /repo/a
- `fs.readdir` does not know /repo/a

So I can not reach staged removing files.

## Real World Case

I'm creating editor and git client (WIP, not OSS yet) https://nervous-kilby-73c9b0.netlify.com/

![](https://i.gyazo.com/0b3cdd87e9b5cbb4d2616f68d9724ca7.gif)

Git client can not define committable stage state.

wmhilton:
Ah! There's been confusion on this topic before (see #206). I'm not 100% sure what you're goal is, but I think you can achieve it by calling `listFiles({ ref: 'HEAD' })`. If a file is deleted from the file system and removed from the git stage, then it is only remembered by the previous commit.

mizchi:
> it is only remembered by the previous commit.

Ah, you are right... I can resolve this issue by `ref: 'HEAD'`

```
await git.listFiles({fs, dir: '/playground'})
(2) ["README.md"]
await git.listFiles({fs, dir: '/playground', ref: 'HEAD'})
(3) ["README.md", "a"]
```

I took `ref` was always `HEAD` for granted.

Thank you!

wmhilton:
> I took ref was always HEAD for granted but it's not so.

Ah, shit you're right! Originally `listFiles` only listed the files in the staging area. Then in #263 I added the `ref` parameter so that listFiles could be used on commits as well. But now it is inconsistent with the rest of the API that uses a `ref` parameter that has a default value of 'HEAD'. Here the default is `undefined` which means it uses the stage... Sigh. I should start making a list of design mistakes that I can try to fix when version 1.0.0 is released.

But hey, at least it's still not as confusing as the git CLI.

Issue #251, opened 2018-05-28
List files with unstaged changes


mofux:
I'm trying to figure out how to retrieve a list of unstaged files that have been modified, similar to the `git status` command. My current approach uses `git.listFiles` to get the list of all files known to the repository, and then calling `git.status` on every single file. Although this works, it is very slow compared to the native `git status` implementation, which returns the results much much faster. Plugging in my own file system implementation, I can see that for every `git.status` call on a file, there is a `readFile .git/index`, followed by `lstat path/to/file/in/repo` followed by `readFile path/to/file/in/repo`.

I think the bottleneck is that every file is read from the filesystem, not sure if that's a bug, because I can see in the `git.status` code that it should only read the file if the attributes returned by the `lstat` command have changed. 

Another bottleneck is that the index file is being read with every single `git.status` request to a file, which could be avoided if the initial index was cached and re-used for subsequent `git.status` calls.

Are there any plans to add an API for this feature, or any hints on how I can optimise my current routine?

wmhilton:
See if the new [statusMatrix](https://isomorphic-git.org/docs/en/statusMatrix) command works for you. I included an example of how to solve for "What files have unstaged changes?" in the docs.

chrisdothtml:
`statusMatrix` works great for me. Is there any existing way to use that `pattern` functionality for `add` (i.e. `git add dir/*`), or is the only way to manually add each file?

brettz9:
Is https://isomorphic-git.org/docs/en/faq#how-to-add-all-untracked-files-with-gitadd outdated now that there is https://isomorphic-git.org/docs/en/statusMatrix#q-what-files-have-unstaged-changes ?

brettz9:
I know the FAQ doesn't add the files, but it is the same concept (or the complete new approach, along with file adding, might also be added to the FAQ).

mhintzke:
@billiegoose  `statusMatrix` works from a functional standpoint, but I am trying ti simply emulate `git reset HEAD` and performing a `statusMatrix` call against a repository with ~2800 files and no changes at all takes 100s to complete on Ubuntu 20.04. This makes it unusable in its current state. I am going to try to test this out on Windows and see if its any faster just to see if it is a simply a platform issue or not.

jcubic:
[...]

Issue #206, opened 2018-05-17
Delete file not listed


jcubic:
## What do I get:

```javascript
await git.listFiles({fs, dir});
// ['foo', 'bar', 'baz'];
await git.remove({fs, dir, 'foo'});
await git.listFiles({fs, dir});
// ['bar', 'baz'];
// should be [['foo', 'bar', 'baz'];
await pfs.readDir(dir);
// ['foo', 'bar', 'baz']
await git.status({fs, dir: '/test', filepath: 'foo'})
// undocumented "*undeleted"
await git.add({fs, dir: '/test', filepath: 'foo'})
await git.status({fs, dir: '/test', filepath: 'foo'})
// "unmodified"

await git.remove({fs, dir, 'foo'});
await pfs.unlink('/test/foo');
await git.status({fs, dir: '/test', filepath: 'foo'})
// deleted which is fine but
await git.listFiles({fs, dir});
// ['bar', 'baz'];
// should be [['foo', 'bar', 'baz'];
```

## What I expect:
the file should always be on the list because it's not yet committed.
When I call `fs.unlink('/test/foo');` the file have correct status deleted but it's not one the list.

jcubic:
I've found this code by try and error:

```javascript
git.remove({fs, dir: '/test', filepath: 'bar'}).then(() => git.add({fs, dir: '/test', filepath: 'bar'})).then(() => fs.unlink('/test/bar'));
```

it make the file appear as deleted but I can't call git.add on that file because I got error file not found.

wmhilton:
`listFiles` lists the files currently in the git index (sometimes called the git stage). I *think* you'll find its behavior is consistent with `git ls-files`... I just tried it and `git ls-files` will not list a file if it has been removed with `git rm`, even though that change is not committed.

wmhilton:
What you *might* want is to list the files from the current HEAD commit. Which, due to the way git stores things internally, is a more complicated recursive task. For more details see [this thread](https://github.com/isomorphic-git/isomorphic-git/issues/43#issuecomment-361017144)

wmhilton:
And if that's what you want, I'll follow through on what I told @mojavelinux and write up the code that does that and add it to the FAQ #191

mojavelinux:
I agree files which are removed should not be shown in listFiles as they are no longer part of the index. This is different from reading from the working tree, where the file still lives, as the API docs for remove states:

> Note that this does NOT delete the file in the working directory.

[...]

PR #355, opened 2018-07-29
Walker

wmhilton:
This PR has a lot to wrap one's head around. But it's awesome. And will be foundational for building `git diff` and `git merge` and probably other things I haven't thought of.

## git.statusMatrix
This is a more efficient way to get the status of all the files in a repo. Instead of the complicated recipes for calling `git.status` on all the files in the workdir + HEAD, you can call this. It returns data in the form of a matrix where the values indicate which copies of a file/blob are identical. Think of the natural progression as being from HEAD (prev) to WORKDIR (current) to STAGE (next). Then HEAD is "version 1", workdir is "version 2", and staged is "version 3". Then, imagine a virtual "version 0" which is completely empty. So for a file to be identical to "version 0" means the file is deleted.

| filename | HEAD | WORKDIR | STAGE | equivalent `status` output |
|---|---|---|---|---|
| a.txt | 1 | 1 | 1 | unmodified |
| b.txt | 1 | 2 | 1 | *modified |
| b2.txt | 1 | 2 | 3 | *modified |
| b3.txt | 1 | 2 | 2 | modified |
| c.txt | 1 | 0 | 1 | *deleted |
| c2.txt | 1 | 0 | 0 | deleted |
| d.txt | 0 | 2 | 0 | *added |
| d2.txt | 0 | 2 | 3 | *added |
| d3.txt | 0 | 2 | 2 | added |
| e.txt | 0 | 0 | 0 | absent |
| weirdo.txt | 1 | 1 | 3 | *unmodified |
| weirdo2.txt | 1 | 1 | 0 | *undeleted |
| weirdo3.txt | 0 | 0 | 3 | *absent |

As you can see, this actually makes it possible to distinguish between a few more combinations than was possible with the text based labels used in `status`. It also makes it possible to quickly answer many types of questions such as:

Q: What files have been deleted? #284 #206
A: Any where WORKDIR === 0

Q: What files have unstaged changes? #251 
A: Any where WORKDIR !== STAGE

Q: What files have been modified since the last commit?
A: Any where HEAD !== WORKDIR

Q: What files will _NOT_ be changed if I commit right now?
A: Any where HEAD === STAGE

- [ ] add TypeScript definition in `src/index.d.ts`
- [ ] add command to README list of commands
- [ ] create a test in `src/__tests__`
- [ ] make a feature commit "feat: Added 'X' command"

## git.walkBeta1
`statusMatrix` is built upon a new `walk` command that I am trying to expose. It simplifies comparing all the filepaths in two or more trees. Trees can be a file directory, a git commit, or a git index (aka stage). So you can compare two file directories, or 10 commits, or the stage of one repo with the working directory of another repo... whatever. As long as a file or directory is present in at least one of the trees, it will be traversed in alphabetical order. The arguments to `walk` are the `trees` you want to traverse, and 4 optional transform functions: `filter`, `map`, `reduce`, and `iterate`.

The trees are represented by three GitWalker magic functions that can be imported: `TREE({fs, gitdir, ref})`, `WORKDIR({fs, dir, gitdir})`, and `STAGE({fs, gitdir})`. These functions return objects that implement the `GitWalker` interface, which is an implementation detail that is hidden from the user in a Symbol property. The only thing they are good for is passing into `walkBeta1`'s `trees` argument.

`filter`, `map`, `reduce`, and `iterate` allow you control the recursive walk by pruning and transform `TreeEntry`s into the desired result.

`filter(TreeEntry[]) => boolean` This is a good place to put limiting logic such as skipping entries with certain filenames. If you return false for directories, then none of the children of that directory will be walked.

`map(TreeEntry[]) => any` This is a good place for query logic, such as calling `if (head.content.toString('utf8').contains('foo')) {`. Ultimately, compare all the entries and return any values you are interested in, such as `return {filename: head.fullpath, oid: head.oid, diff: diff(head.content, workdir.content)}`.

`reduce(mappedParent, mappedChildren[])` The default implementation of this function is a flat map, so all directories and children are returned in a giant flat array. You can be fancy though and compute a hierarchical structure here, such as `(parent, children) => Object.assign(parent, { children })`

`iterate(walk, children)` The default implementation recurses all children concurrently using Promise.all. However you could use a custom function to traverse children serially or use a global queue to throttle recursion.

`TreeEntry` is an interface that abstracts computing many common tree / blob stats.
By default, you get only
```js
{
  fullpath: string;
  basename: string;
  exists: boolean;
}
```
Additional properties are computed upon request. This lets you build lean, mean, efficient walking machines.
```js
await entry.populateStat() // provides
entry.type // 'tree', 'blob'
// and where applicable, these properties:
{
    ctimeSeconds: number;
    ctimeNanoseconds: number;
    mtimeSeconds: number;
    mtimeNanoseconds: number;
    dev: number;
    ino: number;
    mode: number;
    uid: number;
    gid: number;
    size: number;
}

await entry.populateContent() // provides
// except for STAGE which cannot provide content
entry.content // Buffer

await entry.populateHash() // provides
entry.oid // SHA1 string
```

- [ ] add TypeScript definition in `src/index.d.ts`
- [ ] add command to README list of commands
- [ ] create a test in `src/__tests__`
- [ ] make a feature commit "feat: Added 'X' command"


wmhilton:
Alright. I'm very nearly satisfied with statusMatrix. It turned out the `filename` parameter's prefix matching wasn't useful enough; I needed a way to efficiently get the status of only certain files based on their file extension (multiple suffix matching). So after carefully picking the smallest libraries I could find, `statusMatrix` now has a `pattern` parameter that supports glob syntax!

I'll add some more options in the near future - @akaJes suggested in #358 an option to return the full SHA1 values rather than just relative values, which I think is a good option to have. I'll probably also add an `ignored: true` flag to allow listing ignored files.

But this branch has a handful of important bug fixes: some to the way that index files are generated that may be causing #362, and also an alphabetical sorting bug in my readdir implementation that only appears in obscure edge cases and may be platform specific. So I want to get this shipped. Hence, I've dubbed the walker `walkBeta1` 😆.

Documentation will be forthcoming.

PR #356, opened 2018-07-29
fix multibyte filename length

### LINKED ISSUE #354: git.add adds wrong filepath when filepath is multibyte string (like Japanese)
I encountered on Node@10, Chrome 68 and isomorphic-git@0.24.2

```js
  const root = '/my-root'
  const addingFilepath = "日本語"
  await fs.promises.writeFile(path.join(root, addingFilepath), "1")
  await git.add({ fs, dir: root, filepath: addingFilepath })
  await git.commit({
    fs,
    dir: root,
    message: "Add japanese file",
    author: { email: "dummy", name: "dummy" }
  })
  const files = await git.listFiles({ fs, dir: root })
  console.log(files)
```

This result lost some data

```
 [ '日' ] // expected:  ['日本語']
```

bokuweb:
fix #354


TomasHubelbauer:
Does this fix #354? Why did you not use the `Buffer` solution @wmhilton suggested? Does it not work in all cases? The `encodeURIComponent`+regex solution feels weird to me, I feel like it must be expensive? But I may be wrong on this, just wondering.

TomasHubelbauer:
No worries @bokuweb! Silly me for not leaving that as a review comment, but alas. Thank you for the contribution!

wmhilton:
This is awesome! I'm not 100% sure it fixes it though. Can one of you add a test that adds a file named 日本語, commits it, then does listFiles to confirm it didn't get truncated? Similar to the example in #354.

bokuweb:
@wmhilton Thanks for your quick response :) I added a multibyte filename test. Could you please check this, if you would not mind :)

wmhilton:
I'm going to add a test to `commit` as well, because GitTree stores paths differently than GitIndex

PR #363, opened 2018-08-02
feat: Push a local branch/tag to a differently named branch/tag on the remote

### LINKED ISSUE #348: Push a branch/tag to a different branch
Hey,

I would like to push foobar to the remote master branch.
With the command line, I would do this:

```
git push origin foobar:master
```

I would like to do the same with isomorphic-git.
Something like that:

```js
git.push({ fs, dir: '.', ref: 'foobar:master' })
```

```js
git.push({ fs, dir: '.', ref: 'tag-1.2.3:master' })
```

This means we need to "parse" refs but I think there's already something like that in the code.

WDYT?

hsablonniere:
Fixes #348

Just a quick & dirty patch to make this work:

```js
git.push({ ...pushDetails, ref: 'feature-branch:refs/heads/master' }))
```

### DIFF
--- a/src/commands/push.js
+++ b/src/commands/push.js
@@ -60,12 +60,6 @@ export async function push ({
     } else {
       fullRef = await GitRefManager.expand({ fs, gitdir, ref })
     }
-    let fullRemoteRef
-    if (!remoteRef) {
-      fullRemoteRef = fullRef
-    } else {
-      fullRemoteRef = await GitRefManager.expand({ fs, gitdir, ref: remoteRef })
-    }
     let oid = await GitRefManager.resolve({ fs, gitdir, ref: fullRef })
     let auth = { username, password, token, oauth2format }
     let GitRemoteHTTP = GitRemoteManager.getRemoteHelperFor({ url })
@@ -83,7 +77,25 @@ export async function push ({
       finish: httpRemote.refs.values()
     })
     let objects = await listObjects({ fs, gitdir, oids: commits })
-
+    let fullRemoteRef
+    if (!remoteRef) {
+      fullRemoteRef = fullRef
+    } else {
+      try {
+        fullRemoteRef = await GitRefManager.expandAgainstMap({
+          ref: remoteRef,
+          map: httpRemote.refs
+        })
+      } catch (err) {
+        if (err.code === E.ExpandRefError) {
+          // The remote reference doesn't exist yet.
+          // If it is fully specified, use that value. Otherwise, treat it as a branch.
+          fullRemoteRef = remoteRef.startsWith('refs/') ? remoteRef : `refs/heads/${remoteRef}`
+        } else {
+          throw err
+        }
+      }
+    }
     let oldoid =
       httpRemote.refs.get(fullRemoteRef) || '0000000000000000000000000000000000000000'
     if (!force) {
--- a/src/commands/push.js
+++ b/src/commands/push.js
@@ -27,6 +27,7 @@ export async function push ({
   fs: _fs = cores.get(core).get('fs'),
   emitter,
   ref,
+  remoteRef,
   remote = 'origin',
   url,
   force = false,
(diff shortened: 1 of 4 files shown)


hsablonniere:
I just updated the impl. We're progressing...
I have no idea how to test that properly.

@wmhilton WDYT?

hsablonniere:
## TODO:

* [ ] Decide if we want to handle those cases for `<src>`:
  * a SHA-1 id in a push `<src>`
  * a SHA-1 id in a push `<src>:<dst>`
  * a SHA-1 expression in a push `<src>`
  * a SHA-1 expression in a push `<src>:<dst>`
* [ ] Handle errors
* [ ] Update docs

wmhilton:
I think we can do even better, and make the API a little more JavaScripty. I don't particularly like how `ref` now has a regex parser and how supporting the "+" creates a potential conflict with the `force` parameter. (What if a user explicitly sets `force` to false but the refspec contains a "+"?)

What is we just introduce an optional `remoteRef` parameter? That way none of the arguments have the possibility to conflict with each other.

```js
// Instead of 
git.push({ ref: '+master:feature/issue-363' })
// use an array of { src, dest }
git.push({ ref: 'master', remoteRef: 'feature/issue-363', force: true })
```

If omitted, the remote ref is inferred as it currently is, and if present it will use that value instead.

If the application you have in mind currently uses refspecs... you could either parse the refspec with the regex in the client application, or we could introduce a `refspec` parameter and make its usage mutually exclusive with using `ref` and `force`.

wmhilton:
For testing, I have this crazy system. Currently it only works in `karma` because I haven't figure out how to integrate it with `jest`. But I think we'll be able to just add a new test case to [https://github.com/isomorphic-git/isomorphic-git/blob/master/__tests__/test-push.js`](test-push.js) and then inspect the returned `<PushResponse>`.

hsablonniere:
> I think we can do even better, and make the API a little more JavaScripty. I don't particularly like how ref now has a regex parser and how supporting the "+" creates a potential conflict with the force parameter.

I understand. In our use case at Clever Cloud, we only expect users to provide a short branch name and we always want to have a `remoteRef` equal to `'master'` so I don't really need this refspect parsing.

> (What if a user explicitly sets force to false but the refspec contains a "+"?)

I wanted to try with the git client but you only have `--force`, I'm not sure how to do that.

Anyway, let's simplify this! :+1:

hsablonniere:
@wmhilton I updated/simplified the code according to what you said.

src/commands/push.js
```diff
@@ -59,6 +60,12 @@ export async function push ({
     } else {
       fullRef = await GitRefManager.expand({ fs, gitdir, ref })
     }
+    let fullRemoteRef
+    if (!remoteRef) {
+      fullRemoteRef = fullRef
+    } else {
+      fullRemoteRef = await GitRefManager.expand({ fs, gitdir, ref: remoteRef })
```
wmhilton: Technically this may not do what we want, because `.expand` searches through _local_ refs. I think I see how to fix this though.

wmhilton: Ah. Yeah, actually this will crash it if trying to push to a new branch (that doesn't exist yet). I got this.

wmhilton:
Alrighty. I'm 99% satisfied with it. The only ambiguity / unexpected behavior is if you push a local tag (`ref: 'v1.0.0'`) to a new ambiguous name (`remoteRef: 'foobar'`) - it will create `refs/heads/foobar` NOT `refs/tags/foobar` on the remote. 

I originally had it auto-detect if the source was a branch or a tag and use that to infer the destination... but there's no legitimate reason to think you would WANT that.

If I ran  `git.push({ref: 'v1.0.0', remoteRef: 'releases/1.0.0'})` I should expect that to be equal to `git.push({ref: 'refs/tags/v1.0.0', remoteRef: 'refs/heads/releases/1.0.0'})` not `git.push({ref: 'refs/tags/v1.0.0', remoteRef: 'refs/tags/releases/1.0.0'})`.

It's much easier to simply remember "_tags are weird_" and live by a simple "better put 'refs/tags/' in front" rule-of-thumb than think "hmm, well is the source type the same or different from the destination type?"

PR #381, opened 2018-08-15
new improved command `reset`

akaJes:
## I'm adding a new command:

- [x] add TypeScript definition in `src/index.d.ts`
- [x] add command to README list of commands
- [x] make a feature commit "feat: Added 'X' command"


wmhilton:
I'd pay good money for a bot that just runs `standard --fix` on incoming pull requests. I like linters, but I hate burdening contributors with running them.

akaJes:
i did it from browser and the best way use travis for errors checking
besides it is only draft

wmhilton:
Should we call it `resetIndex` instead of just `reset`? Simply because `git reset [--soft | --hard | --mixed]` is one of the most confusing command names. 

Alternatively... would it make more sense to build this functionality into the `add` command? E.g.

```js
// updates the index entry for README.md with with the working directory value
await git.add({ fs, dir, filepath: 'README.md' })
```

```js
// updates the index entry for README.md with with the value from any commit, not just HEAD
await git.add({ fs, dir, filepath: 'README.md', ref: 'HEAD' })
```

akaJes:
> Alternatively... would it make more sense to build this functionality into the add command? E.g.

but the native git hasn't this ability

there is one issue with symbolic links - these will work unproperly,
Your FileSystem doesn't support reading symbolic links eg `fs.readlink`

src/commands/resetIndex.js
```diff
@@ -0,0 +1,65 @@
...
+export async function resetIndex ({
+  dir,
+  gitdir = path.join(dir, '.git'),
+  fs: _fs,
+  filepath,
+  ref = 'HEAD'
+}) {
+  try {
+    const fs = new FileSystem(_fs)
+    const tree = await log({ gitdir, fs, ref, depth: 1 })
```
wmhilton: Can you use `src/utils/resolveTree.js` here instead of `log`? `log` will only resolve commits, whereas `resolveTree` will also handle raw OIDs or tags that point directly to `tree` objects.

src/commands/resetIndex.js
```diff
@@ -0,0 +1,65 @@
...
+    let oid = tree[0] && tree[0].tree
+    let workdirOid
+    if (oid) {
```
wmhilton: I suggest instead, `if (!oid) throw new GitError(E.ResolveRefError, {ref})`

src/commands/resetIndex.js
```diff
@@ -0,0 +1,65 @@
...
+        oid = obj && obj.oid
+      } catch (e) {
+        oid = null
+      }
+      const object = await fs.read(path.join(dir, filepath))
+      workdirOid = await GitObjectManager.hash({
+        gitdir,
+        type: 'blob',
+        object
+      })
```
wmhilton: Do wrap `workdirOid = ...` in a `if (object) { ... }` because object can be null if file is missing!

src/commands/resetIndex.js
```diff
@@ -0,0 +1,65 @@
...
+              if (err.code !== 'ENOENT') {
+                throw err
+              }
+            }
+          }
+          index.insert({ filepath, stats, oid })
+        }
+      }
+    )
+    // TODO: return oid?
```
wmhilton: You can delete this TODO 😁

src/commands/resetIndex.js
akaJes: I have only `filepath` and `ref` but `src/utils/resolveTree.js` takes only `oid`

src/commands/resetIndex.js
akaJes: `oid` can be undefined when `ref` not exists or repository is empty

src/commands/resetIndex.js
akaJes: agree

src/commands/resetIndex.js
akaJes: the `log` command can be replaced with this
```
let oid = await GitRefManager.resolve({ fs, gitdir, ref })
const tree = await logCommit({ fs, gitdir, oid })
oid = tree && tree.tree
```

src/commands/resetIndex.js
akaJes: is this will be `null` or will be an exception?

src/commands/resetIndex.js
wmhilton: I say if `ref` is not accessible it should throw this error. 😁

src/commands/resetIndex.js
wmhilton: Actually... you just need the first line. `readObject` uses `resolveTree` internally so all you need is the `oid` of the commit, you don't need to do `logCommit` and get the tree.

src/commands/resetIndex.js
akaJes: When the repository is empty `ref` will be accessible?

akaJes: Besides it depends to the `log` command more, but `theGitRefManager.resolve` can return only an exception

wmhilton:
Thank you @akaJes! You put a lot of work into this! 🎉

Also, it should work with symlinks now. 😁

Issue #386, opened 2018-08-24
http auth not support Digest


denghui1010:
Some git server use Digest auth, but isomorphic-git only support Basic auth. so when clone from the server which use Digest auth will return 401 Unauthorized

wmhilton:
Note to self: To implement this, probably the easiest thing to do would be to try using the [`digest-header`](https://www.npmjs.com/package/digest-header) package and if that adversely affects bundle size, swap out `utility` with a less bloated MD5sum operation. https://bundlephobia.com/result?p=digest-header@0.0.1

wmhilton:
@denghui1010 You can now replace the default HTTP request function with a custom one! Check out the [`http` plugin docs](https://isomorphic-git.org/docs/en/plugin_http). I think this will make it easy to solve this now. If you do end up making an example `http` plugin that handles Digest Authentication, please share it on the community [code snippets page](https://isomorphic-git.org/docs/en/snippets). As always I am more than happy to help!

Issue #202, opened 2018-05-17
support 'agent' for corporate SSL proxies


privatenumber:
I get a certificate error due to the github repo being on an enterprise server:
```
(node:14790) UnhandledPromiseRejectionWarning: Error: unable to get local issuer certificate
    at TLSSocket.onConnectSecure (_tls_wrap.js:1046:34)
    at TLSSocket.emit (events.js:180:13)
    at TLSSocket._finishInit (_tls_wrap.js:633:8)
(node:14790) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(
```

Using `process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';` helps but putting this out there so that there can be a more official way to handle this.

wmhilton:
Are you running it behind a corporate proxy server?

hirokiosame:
Yep, the enterprise is behind a self-signed server hence the certificate issue.

Wanted to share this scenario in case you guys wanted to support passing in certs or an option like `strictSSL` in [request](https://github.com/request/request).

wmhilton:
Ah yes. I had to have a lot of npm modules to work with the corporate proxy when I was doing work for GE. I'm delegating all the actual requests to [simple-get](https://npm.im/simple-get), which I see does support an 'agent' option that can be used with, say, [https-proxy-agent](https://npm.im/https-proxy-agent). We could add an 'agent' option to all the commands that makes requests and pass that option through. Or once the plugin API is finished, we could implement it as a plugin that's initialized once.

wmhilton:
@hirokiosame You can now replace the default HTTP request function with a custom one! Check out the [`http` plugin docs](https://isomorphic-git.org/docs/en/plugin_http). I think this will make it easy to solve this now. If you do end up making an example `http` plugin that can navigate SSL proxies, please share it on the community [code snippets page](https://isomorphic-git.org/docs/en/snippets). As always I am more than happy to help!

PR #393, opened 2018-08-26
[WIP] 'httpAgent' plugin

wmhilton:
Example usage:
```js
const fs = require('fs')
const ProxyAgent = require('proxy-agent')
const {fetch, plugins} = require('.')
plugins.set('fs', fs)
plugins.set('httpAgent', new ProxyAgent('http://localhost:8080'))

fetch({ dir: '.'}).then(console.log)
```

I'm not sure if this works yet though - I still get self-signed cert errors.

I want to see if this can solve both #202 and #386


wmhilton:
This has been superceded by #605.

PR #396, opened 2018-08-26
Credential Manager plugin API

wmhilton:
In order to support submodules we have to recursively clone multiple repos, but those repos might not all use the same username/password. And we don't know what those submodule repo URLs are until we've cloned the first repo! So providing the username and password at the beginning of the clone command is not ideal. Much better would be a way to _dynamically_ request credentials on demand.

This also leads to a better experience for the user. For instance, instead of entering their username/password every time, it can be stored and remembered in a keyring. The username/password combo can be stored for an entire domain name, such as one username/password for `gitlab.com` and another for `github.com` so users almost never need to re-enter their credentials once saved. But, if for some reason the saved password doesn't work (such as when they changed their password), instead of failing `isomorphic-git` could interactively prompt the user for the new password.

@jcubic and @mizchi would you be willing to collaborate on this? I think your Git Web Terminal and Next Editor projects are perfect testing grounds for proving out the credential manager API.

## ☐ credential manager plugin
This API is modeled after the canonical [git-credential](https://git-scm.com/docs/git-credential) API, although I've modified the `fill|approve|reject` to be `fill|approved|rejected` because it took me forever to figure out that "approve" was not _asking_ for approval, but _informing_ the backend that these credentials _had_ been proved successful. (Similarly 'reject' was not telling the backend to reject the credentials, but informing the backend that those credentials _were_ rejected by the server.) Only a single credential provider is registered, BUT since the API is very simple you could easily write a wrapper that combines multiple credential providers and dispatches them however you want.

  - async fill ({ url: string }) -> Promise<auth: JSON>
  - async approved ({ url: string, auth: JSON }) -> void
  - async rejected ({ url: string, auth: JSON }) -> void

Here `auth` can be any combination of `username`, `password`, `token`, and `oauth2format`. (See [authentication](https://isomorphic-git.org/docs/en/authentication) for details.) You can also include additional properties that may only be understood by certain **remote providers** that understand them, such as an `ssh_keyfile` for the hypothetical SSH remote provider. The only requirement is it be a JSON-serializable value.

If during `fill`, the credential helper does not have (or chooses not to provide) any auth information for that url, it should return an empty JSON object.

_Note: whether the path is should be considered when searching for matching credentials _will be left to credentialManager implementations_ but to match `git`'s behavior implementations should respect the config value `useHttpPath`. But let's be honest - that's not the most important detail.


wmhilton:
Here's a strawman credential manager implementation using `localStorage` for persistent storage and `window.prompt` for the UX.

```js
// Completely untested, written in Github comment box lol
const urlify = (protocol, host, path) => `${protocol}://${host}/${path}`
const CredentialManager = {
  async fill ({ protocol, host, path }) {
    let auth = localStorage.get(urlify(protocol, host, path))
    if (!auth) {
      auth = {
        username: window.prompt('Enter username'),
        password: window.prompt('Enter password')
      }
    }
    return auth
  }
  async approved({ protocol, host, path, auth }) {
    if (window.confirm('Do you want to save this password?')) {
      localStorage.set(urlify(protocol, host, path), auth)
    }
  }
  async rejected({ protocol, host, path }) {
    let url = urlify(protocol, host, path)
    if (localStorage.has(url)) {
      if (window.confirm('Authentication was unsuccessful. Delete saved password?')) {
        localStorage.delete(url)
      }
    }
  }
}

git.plugins.set('credentialManager', CredentialManager)
```

wmhilton:
Note that CredentialManagers could also choose to save only the 'username' and prompt for 'password', or could save multiple usernames for a domain (if I have two Github accounts for instance) and prompt the user to choose which one to use.

wmhilton:
@jcubic @mizchi I made an example CredentialManager implementation here: https://github.com/isomorphic-git/credential-manager-node-cli

jcubic:
what is `approved({ protocol, host, path, auth }) {`? is this function with anonymous arrow function as argument? because parenthesis don't match and there are no `=>`.

wmhilton:
Here's a demonstration of a CLI style Credential Manager (namely the WIP [`@isomorphic-git/credential-manager-node-cli`](https://github.com/isomorphic-git/credential-manager-node-cli)) in action. Instead of specifying `--username` and `--password` as CLI arguments, it interactively prompts for them.

capture4

jcubic:
I was confused by shorthand method declaration.

wmhilton:
> I was confused by shorthand method declaration with async.

To be honest, as I remember when I was writing it I thought "wait... will this [shorthand method declaration] still work with the async keyword in front of it?" but lo, it does

wmhilton:
> I have pause in push that I'm testing (my code is bit more complicated because I use web worker) that hide the prompt and in credential plugin I'm ask for input (first prompt is not visible I need to press CTRL+D).

I am very confused. Can I see your plugin code?

wmhilton:
The intention with the credentialManager API is that it could hook into your current `git login` code. Since you already have an entire system in place that stores credentials in localStorage etc.

Since git is in a worker, you'd need to proxy it from the worker to the main thread to get access to the jquery terminal, but since it's a stateless API with pure JSON arguments that should be easier to do.

PR #429, opened 2018-09-03
feat: Added 'detailed' parameter to listFiles command

akaJes:
sugest to add a small option for more information (draft not tested yet)

## I'm adding a parameter to an existing command:

- [x] add parameter to the function in `src/commands/X.js`
- [ ] add parameter to [docs](https://github.com/isomorphic-git/isomorphic-git.github.io/tree/source/docs)/X.md
- [ ] add parameter to the TypeScript library definition for X in `src/index.d.ts`
- [x] make a feature commit "feat: Added 'bar' parameter to X command"


wmhilton:
Haha! @akaJes, don't mind the robot I just wrote him yesterday - he still is in alpha testing.

I like this idea. What kinds of details should `detailed` mode provide?

akaJes:
Look at my first commit - there is only two kinds: the list as default and array of objects.

wmhilton:
@akaJes the objects that it returns are straight from the `GitIndex` implementation... mostly contain `fs.Stat` information but also things like "flags.stage" and "flags.assumeValid" and I don't think I want to let that stuff into the public API without more tests and documentation.

`listFiles` was meant to be an equivalent to [`git ls-files`](https://git-scm.com/docs/git-ls-files) which is a pretty simple command that doesn't have a detailed mode. Also, having the output change from `Promise<string[]>` to `Promise<Object[]>` would make the function harder to define in TypeScript.

I'm guessing you're interested in either the `oid`, `size`, or `mode` of the files. Instead of accepting this PR as is, can we maybe use the more powerful [`walkBeta1`](https://github.com/isomorphic-git/isomorphic-git/pull/355) API to achieve your goal? If you describe your goal I can probably whip up some example code showing how to achieve it using the `walk` API.

wmhilton:
Ignore the previous `walkBeta1` link. I've added some documentation to the website with the [walkBeta1 API](https://isomorphic-git.org/docs/en/walkBeta1)

wmhilton:
@akaJes The more I think about it, the more I like this idea. Can we narrow down the API and specifically expose the "useful" details (oid, mode, size, etc) but not all of them? (inode, dev, etc)?

PR #470, opened 2018-09-19
Clients MUST ignore an LF at the end of the line

DeltaEvo:
## I'm fixing a bug or typo

- [x] make a fix commit "fix: [Description of fix]"

It's specified by the protocol here https://github.com/git/git/blob/master/Documentation/technical/http-protocol.txt#L267
For example bitbucket doesn't send LF :(


DeltaEvo:
Thanks

PR #471, opened 2018-09-19
Get Azure Test Results and Code Coverage pages working

wmhilton:
We can potentially ditch CodeCov (which hasn't handled my force pushes well... I don't think I've seen a CodeCov report that worked in months) and Travis (which has a concurrency of 1 vs Azure's concurrency of 10(!!!)) AND have a nice UI report for all the tests so we don't have to scroll through the enormous log looking for what tests failed.


wmhilton:
Close. Just waitin' to hear back what the equivalent of TRAVIS_PULL_REQUEST_SHA is.

PR #491, opened 2018-09-27
[WIP] don't crash when pushing commits referencing submodules

### LINKED ISSUE #468: Support pushing repos with submodules
Hey,

We use isomorphic-git in our CLI tool [clever-tools](https://github.com/CleverCloud/clever-tools) and when we try to push a project with submodules it fails.

I published 2 very simple repos so you can reproduce the problem:

* https://github.com/hsablonniere/clever-test
* https://github.com/hsablonniere/clever-test-submodule

Here is the situation:

```sh
# on the clever-test repo
git log
* faa7e78 add the-submodule
* 47c8612 first commit
```

```sh
# on the clever-test repo
git diff HEAD~
# ...
+Subproject commit fda77d789934f2896b5a3f5735e7df9e690c90be
```

```sh
# on the clever-test repo
cat .git/modules/the-submodule/refs/heads/master
fda77d789934f2896b5a3f5735e7df9e690c90be
```

```sh
# on the clever-test-submodule repo
git log
* fda77d7 cccc
* 5bce34a bbbb
* 07eafb4 aaaa
```

When we do a clever deploy (which internally calls the `push()` function), we have this:

```sh
# on the clever-test repo
clever deploy
Pushing source code to Clever Cloud.
(node:27425) UnhandledPromiseRejectionWarning: ReadObjectFail: Failed to read git object with oid fda77d789934f2896b5a3f5735e7df9e690c90be
    at readObject (/snapshot/clever-tools_1.0.0-beta.22-N2XRLBNNUCCIYIWSUQLZSGXTQQ6RO7KTCEI4CQS3OTIJFHLQVX7A/node_modules/isomorphic-git/dist/for-node/isomorphic-git/index.js:3005:11)
```

isomorphic-git tries to read the `fda77d789934f2896b5a3f5735e7df9e690c90be` commit in the index. Since this commit comes from the submodule repo, it fails.

We need isomorphic-git to read the list of submodules and be able to recursively read their indexes.

wmhilton:
might resolve #468

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -43,10 +43,14 @@ steps:
     summaryFileLocation: '$(System.DefaultWorkingDirectory)/coverage/cobertura-coverage.xml'
     reportDirectory: '$(System.DefaultWorkingDirectory)/coverage/lcov-report'
 
+- script: |
+    npm pack
+  displayName: 'npm pack'
+
 - task: PublishBuildArtifacts@1
   inputs:
-    artifactName: 'dist'
-    PathtoPublish: '$(System.DefaultWorkingDirectory)/dist'
+    artifactName: 'npm-tarball.tgz'
+    PathtoPublish: '$(System.DefaultWorkingDirectory)/isomorphic-git-0.0.0-development.tgz'
 
 - task: PublishBuildArtifacts@1
   inputs:
--- a/src/commands/listObjects.js
+++ b/src/commands/listObjects.js
@@ -28,7 +28,11 @@ export async function listObjects ({
     } else if (type === 'tree') {
       let tree = GitTree.from(object)
       for (let entry of tree) {
-        visited.add(entry.oid)
+        // only add blobs and trees to the set,
+        // skipping over submodules whose type is 'commit'
+        if (entry.type === 'blob' || entry.type === 'tree') {
+          visited.add(entry.oid)
+        }
         // only recurse for trees
         if (entry.type === 'tree') {
           await walk(entry.oid)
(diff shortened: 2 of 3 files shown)


wmhilton:
@hsablonniere I'll add some tests tomorrow, but this _ought_ to do the trick. At least it should allow pushing a superproject (yes, that's the actual [terminology](https://git-scm.com/docs/gitsubmodules)!) without naive `readObject` errors.

hsablonniere:
@wmhilton Hey William, this is great. I tried it on several projects and it works! I managed to push repos with submodules.

I have one problem though. The `listObject` function seems to be really slow.
On one of my projects it took around 65 seconds, just to do this step :scream:

I had a quick look at the code and how the recursion is done. In the case of my project, the `walk` function is called `201730` times while only returning `14947` objects. I tried to add this at the beginning and it helped a lot :tada: 

```js
// ...
async function walk (oid) {
  if (visited.has(oid)) {
    return
  }
  visited.add(oid)
// ...
```

It was a very naive try but it seems to work.
Let me know what you think of this and if it has bad side effects I was not aware of.

hsablonniere:
OK, so I'm probably missing the point of this since I have those results:

Before the change:

* listObjects in 68s
* visited 14947 oids
* walk was called 201731 times

After the change:

* listObjects in 2s
* visited 5545 oids
* walk was called 15962 times

I'm confused.

hsablonniere:
So the "problem" I detected once I tried to push my "big" project with submodules is actually related to `listCommits` dealing with forced push. I'll create a separate issue.

hsablonniere:
To be clear: I confirm this commit/branch fixes the pushing project with submodule ;-) Well done @wmhilton

wmhilton:
Awesome! Thanks for testing it @hsablonniere !

Issue #520, opened 2018-10-12
Allow using annotated tags as ref when pusing code


hsablonniere:
Right now, if you use an annotated tag as a ref like this: `push({ ref: '1.0.0', remoteRef: 'master' })`, it will fail.

In [the code](https://github.com/isomorphic-git/isomorphic-git/blob/master/src/commands/push.js#L67) of the `push` command, there is a line where we `resolve` the ref to an oid:

```
let oid = await GitRefManager.resolve({ fs, gitdir, ref: fullRef, peel: true })
```

If the `fullRef` is an annotated tag, the `oid` will be the object id of the tag object itself and not the associated commit SHA1.

A bit like I proposed in #519, I would like `push()` to peel annotated tags when retrieving their commit SHA1. If we expose a `peelTags` option on the `expandRef()` command, we'll have to implement the option on the internal `GitRefManager.resolve()` which means it would be accessible to the push command.

WDYT?

wmhilton:
Good catch! Will fix this weekend.

wmhilton:
This was fixed in #522 and I just forgot to close this.

Issue #519, opened 2018-10-12
Provide a way to resolve commit SHA1 from tag refs


hsablonniere:
Right now, `resolveRef({ ref: '1.0.0' })` will not return the commit SHA1 associated to the tag `1.0.0` but the object id of the tag.

I need a way to use `resolveRef` and retrieve the commit SHA1 if it's a tag. I think I saw somewhere that this process is called peeling.

Something like this would be awesome:

```js
const tagObjectId = await resolveRef({ ref: '1.0.0' })
const tagCommitId = await resolveRef({ ref: '1.0.0', peelTag: true })
```

WDYT?

wmhilton:
I kind of like the purity of `resolveRef`. As you saw in your PR, once you try resolving things recursively, then the `resolveRef` function suddenly has to pull in the `readObject` function and the parser for the `GitAnnotatedCommit` objects... and I'd hate for a bundled web app that just needs to resolve `HEAD` to an oid to have to drag in all that extra code.

So for better tree-shaking / code-spitting, what do you think about making this a new function called `resolveCommit`? We can also probably safely expose `utils/resolveTree`.

[Edit: here I start to ramble a bit, but it's related]

I've been thinking about making lighter-weight, concise, safer alternatives to `readObject()` as well. Currently, I have a lot of code that looks like this:
```js
    let { type, object } = await readObject({ fs, gitdir, oid, format: 'parsed' })
    if (type !== 'commit') {
      throw new GitError(E.ObjectTypeAssertionFail, {
        oid,
        expected: 'commit',
        type
      })
    }
```

and I've been thinking since there are only 4 kinds of objects in git, it wouldn't be terrible to make four different functions: `readCommit`, `readTag`, `readTree`, and `readBlob` that would assert the type is correct and handle throwing an error if necessary, which makes the code much cleaner to read:
```js
    let object = await readCommit({ fs, gitdir, oid })
```

This also is good for code-splitting/tree-shaking. If you include `readObject` in your code right now, you drag in all the parsers for every kind of git object. If you only include `readCommit`, then only the commit parser would be bundled in your application.

So if we try to harmonize these two ideas, we get 8 new functions:

These ones, which will take in a ref or oid (and in the case of tree and blob, take in a filepath) and will peel tags (or dig into sub-trees based on a filepath) to return an `oid`:
- `resolveTag`
- `resolveCommit`
- `resolveTree`
- `resolveBlob`

[...]

PR #521, opened 2018-10-12
WIP peel tags

hsablonniere:
I was trying to understand the details of #519 & #520 and I have a working solution. It's WIP, it lacks tests but let me know what you think of the approach.


wmhilton:
There is (finally) a decent-ish API for peeling tags: https://github.com/isomorphic-git/isomorphic-git/issues/519#issuecomment-573460281

Issue #597, opened 2018-11-29
Error when launching through angular


roddone:
Hi, 

I'm using isomorphic-git in browser through an Angular 7 project. I successfully installed the project with `npm install --save isomorphic-git` but when launching my project with `ng serve` I encouter the following error : 

`ERROR in node_modules/isomorphic-git/src/index.d.ts(13,11): error TS2580: Cannot find name 'Buffer'. Do you need to install type definitions for node? Try `npm i @types/node`.
node_modules/isomorphic-git/src/index.d.ts(115,13): error TS2580: Cannot find name 'Buffer'. Do you need to install type definitions for node? Try `npm i @types/node`.node_modules/isomorphic-git/src/index.d.ts(561,11): error TS2580: Cannot find name 'Buffer'. Do you need to install type definitions for node? Try `npm i @types/node`.`

`@types/node` is already installed and I tried updating it does not solve my problem.

The error is thrown when I import isomorphic-git in my components : `import { clone } from 'isomorphic-git'`

Is it a known problem and do you know if the library is compatible with angular projects ?

Regards, 

Romain.

wmhilton:
Darn. I really have no idea... I'm just googling "Typescript TS2580" but I found this: https://github.com/aws/aws-sdk-js/issues/1271

Maybe that'll help? Apparently there's a setting in a file called "`tsconfig.app.json`" that needs to be turned on for it to see `@types/node` in the client (browser) code.

wmhilton:
Good news it seems to work with Angular 7 in this online sandbox: https://stackblitz.com/edit/angular-pdbaet

roddone:
It works in the sandbox but I tried to download the project and ran it into my local dev environment and I got the same errors as above. I assume the sandbox has a different build process.

It seems isomorphic-git package requires node libraries that are note compatibles with angular because they are "Node" packages : https://github.com/angular/angular-cli/issues/11130
My team and I tried several workarounds but none of them worked because everytime isomorphic package is loaded, it references one of these "node" packages and it kills angular build process.

wmhilton:
> It works in the sandbox but I tried to download the project and ran it into my local dev environment and I got the same errors as above. I assume the sandbox has a different build process.

Hmm you're right! That is frustrating.

This appears to be a popular frustration in the Angular community: https://github.com/angular/angular-cli/issues/9827

I've tried adding Buffer to `src/polyfills.ts` but that didn't seem to work either.

[...]

PR #604, opened 2018-12-01
feat: remove 'path' dependency

wmhilton:
You're probably wondering: why hate on the 'path' module? Well.... see #597. It boils down to: relying on a "builtin" path module makes us dependent on whatever shims builders ship with. Webpack I think is _still_ using an older version of `path-browserify` that is missing `path.posix`. The mere fact that the `path` in the browser was different from the `path` module for node was a source of headaches, because I'd write code that worked great in Jest, then discover it was broken in the browser.

Reading the rationale behind Angular's decision to stop polyfilling node builtins made me reconsider the dangers of relying on them. Although this library doesn't currently get tested with anything but Node and Webpack... I'd love to support Nashorn and [XS](http://www.moddable.com/) and everything under the sun. The more the code relies on polyfills and builtins, the harder that becomes.

Lastly... have you _read_ the source code for the `path` module? It's insane! `path-browserify` is 6kb before gzip! It contributes as much the the bundle size as the SHA1 implementation! And this for a module that I basically only use for one function, `path.join`.

image


karma-pr-reporter:
## Test Results for 416621a2ac6f2423c6e2458e47c9132be56fd66d:
| Browser | Passed | Skipped | Failed | Time | Disconnected | Error |
|---------|--------|---------|--------|------|--------------|-------|
| HeadlessChrome 0.0.0 (Linux 0.0.0) | 222 | 1 | 0 | 59.177 secs | false | undefined |
| Firefox 62.0.0 (Ubuntu 0.0.0) | 222 | 1 | 0 | 1 min 36.029 secs | false | undefined |
| Safari 12.0.0 (Mac OS X 10.13.6) | 222 | 1 | 0 | 2 mins 33.092 secs | false | undefined |
| Chrome Mobile 67.0.3396 (Android 7.1.1) | 222 | 1 | 0 | 3 mins 19.445 secs | false | undefined |
| Edge 17.17134.0 (Windows 10 0.0.0) | 222 | 1 | 0 | 3 mins 53.721 secs | false | undefined |
| Mobile Safari 11.0.0 (iOS 11.2.0) | 222 | 1 | 0 | 4 mins 29.66 secs | false | undefined |

PR #925, opened 2019-11-11
No Buffer dependency [WIP]

wmhilton:
BREAKING CHANGE: methods will return `Uint8Array`s instead of `Buffer`s.

Reasoning:
- `Buffer` is not isomorphic, and using it breaks Angular. (You can polyfill it manually but it doesn't work out-of-the-box.)
- Why should we be using `Buffer` when `Uint8Array` has been around forever and is part of the language.
- This let's us remove the `buffer` module (which gets automatically inserted by Webpack) shaving off a whopping 5.71 KB.

I've confirmed that `sha.js` is the only module using `Buffer` (via `safe-buffer`). Since we recently enabled WebCrypto SHA1 as an optional optimization, and in Node we could always use the native SHA1... I definitely wonder if we could get away with dropping the `sha.js` dependency. That would save an additional 1 KB, and when I compile that with Webpack, we now are at 64 KB instead of 71 KB gzipped, or a 10% reduction, which is pretty phenomenal. The only bigger improvement would be if we could drop packo.


nadilas:
Hi @wmhilton,

> This lets us remove the buffer module (which gets automatically inserted by Webpack) shaving off a whopping 5.71 KB.

this would also enable vercel's edge runtime to work:

> error - Error: The edge runtime does not support Node.js 'buffer' module.                        
> Learn More: https://nextjs.org/docs/messages/node-module-in-edge-runtime

jcubic:
@nadilas I'm afraid this will never be finished. OP is not working on the project anymore.

nadilas:
thanks @jcubic, does that mean it is in general not recommended to use the library at all?

jcubic:
@nadilas No, it means that the project is driven only by contributors, it's maintained but there is no single person that does all the hard work of adding new features and fixing bugs. If you want to help and remove Buffer dependency you are more than welcome to do this.

nadilas:
@jcubic as it should be. I'll take a crack at it

PR #1172, opened 2020-07-01
feat: git log the history of one file only

snowyu:
Currently the `git.log` shows all history of the commits. Miss the commits history of the filepath only.

```bash
git log -- [filepath]
```


snowyu:
with follow supports now, you should set `follow` and `force` both to true.

snowyu:
The another same pr is [snowyu:feat/log-async-iterator](https://github.com/snowyu/isomorphic-git/tree/feat/log-async-iterator), It make code more clear to read for using the async iterator. But I have to add the `@babel/plugin-proposal-async-generator-functions` plugin.

wmhilton:
Regarding: `src/utils/resolveFileIdInTree.js`

This is tricky to do well without having implemented git's actual algorithm for detecting file renames. Comparing just the file `oid` and grabbing the first one that matches a top-down breadth-first search seems risky. What about using `_walk` and grabbing all the blob paths that match `oid` and only following if there is 1 and exactly 1 match? That way we won't accidentally follow the wrong file.

It ends up doing more CPU/IO work to exhaustively determine the file moved rather than was copied, but `_walk` is also optimized for concurrency so it might even be faster than the current algorithm in some case.

snowyu:
Aha, I missed a plurality of the same content exists, thank you pointed out.

The `_walk` is too special(advanced) to walk `GitTree`.  We need a low-level `walkTree` which walks GitTree only.

snowyu:
Ok, I've resolved though comparing all the filepaths with last all filepaths of the same content to determine the correct file name.

jcubic:
Hi, I'm a new maintainer and I will be able to merge your PR if you resolve the conflicts in the files.

snowyu:
Merged.

src/api/log.js
```diff
@@ -13,9 +13,12 @@ import { join } from '../utils/join.js'
  * @param {FsClient} args.fs - a file system client
  * @param {string} [args.dir] - The [working tree](dir-vs-gitdir.md) directory path
  * @param {string} [args.gitdir=join(dir,'.git')] - [required] The [git directory](dir-vs-gitdir.md) path
+ * @param {string=} args.filepath optional get the commit for the filepath only
```
jcubic, line 16: We decided that there should be single parameter `string | string[]`

src/api/log.js
```diff
@@ -13,9 +13,12 @@ import { join } from '../utils/join.js'
...
  * @param {string} [args.ref = 'HEAD'] - The commit to begin walking backwards through the history from
  * @param {number} [args.depth] - Limit the number of commits returned. No limit by default.
  * @param {Date} [args.since] - Return history newer than the given date. Can be combined with `depth` to get whichever is shorter.
+ * @param {boolean=} args.force do not throw error if filepath is not exist (works only for a single file). defaults to false
```
jcubic, line 20: Please fix the docstring `[args.force=false]`

src/api/log.js
```diff
@@ -13,9 +13,12 @@ import { join } from '../utils/join.js'
...
+ * @param {boolean=} args.follow Continue listing the history of a file beyond renames (works only for a single file). defaults to false
```
jcubic, line 21: Please fix the docstring `[args.follow=false]`

jcubic (changes requested):
Please check the conflict, there was a change:
```javascript
    if (tips.length === 0) {
      break
    }
````
was removed. 

Also please remove the example git hooks from testing fixtures.

**Comment by snowyu** [comment]
It has changed to this: 

https://github.com/snowyu/isomorphic-git/blob/634e7da134f96edd1e5b1f34c116bd024f4e219f/src/commands/log.js#L161

```js
    // Stop the loop if there are no more commit parents
    if (tips.length === 0) {
      endCommit(commit)
      break
    }
```

src/api/log.js
snowyu: multi filepaths? maybe should add another `log-files` api to keep the returned result unchanged.

snowyu: Or do your mean to insert the all commits of filepaths to the result? Manually indicate which is the change history of the same file with different file names.

jcubic: Sorry my mistake, I'm reading too many issues at once. That change with filepath was for git.add in #1228

jcubic:
You added a line to the code that got removed. There are no longer that  `if (tips.length === 0) {` check in the code.
```javascript
    // Stop the loop if there are no more commit parents
    if (tips.length === 0) {
      endCommit(commit) // this make the conflict because surrounding code got removed
      break
    }
```

jcubic:
So please update your fork with the latest version. And check how you can do the same without that line. Right now the code don't use `while (true) {` and break anymore.

snowyu:
Oh, I didn't notice that the mian branch has been updated again. done..

jcubic:
Note that `{Boolean=}` was correct the option is optional. You should not remove that only added default parameter and name in bracket.

jcubic:
You've merged `origin/main` but not updated your fork with `upstream/main`.

jcubic:
Sorry I was mistaken your fork was updated. I need to look closer at your code since the logic of that loop was modified recently.

jcubic (approved):

Issue #1124, opened 2020-05-08
TypeError: Cannot read property 'index' of undefined


linfaxin:
TypeError: Cannot read property 'index' of undefined
    at Function.acquire (src/managers/GitIndexManager.js:49)
    at _commit (src/commands/commit.js:62)
    at _merge (src/commands/merge.js:140)
    at async _pull (src/commands/pull.js:96)
    at async Module.pull

The "cache" param may miss when call `_merge()` in `_pull(src/commands/pull.js:96)`.

linonetwo:
This is also causing

```
[1] (node:31230) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'index' of undefined
[1]     at Function.acquire (/Users/linonetwo/Desktop/repo/TiddlyGit-Desktop/node_modules/isomorphic-git/index.cjs:842:16)
[1]     at _commit (/Users/linonetwo/Desktop/repo/TiddlyGit-Desktop/node_modules/isomorphic-git/index.cjs:4338:26)
[1]     at _merge (/Users/linonetwo/Desktop/repo/TiddlyGit-Desktop/node_modules/isomorphic-git/index.cjs:8445:23)
[1]     at async _pull (/Users/linonetwo/Desktop/repo/TiddlyGit-Desktop/node_modules/isomorphic-git/index.cjs:8556:5)
```

wmhilton:
Ah, #1189 mispelled "fixes" as "fixs" so the release bot didn't post it here but I'll post it:

:tada: This issue should be fixed in version 1.7.1 :tada:

The release is available on:
- [npm package (@latest dist-tag)](https://www.npmjs.com/package/isomorphic-git/v/1.7.1)
- [GitHub release](https://github.com/isomorphic-git/isomorphic-git/releases/tag/v1.7.1)

Your **[semantic-release](https://github.com/semantic-release/semantic-release)** bot :package::rocket:

linonetwo:
OMG I've used isomorphic-git before... For some reason I forget I switched to dugite for a year, but due to https://github.com/desktop/dugite/issues/440 I decided to switch back here...

PR #1189, opened 2020-07-11
fix: Cannot read property 'index' of undefined when "Pull"

linonetwo:
## I'm fixing a bug or typo

This fixs https://github.com/isomorphic-git/isomorphic-git/issues/1124

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -425,6 +425,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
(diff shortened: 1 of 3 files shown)


wmhilton:
Wow you're right! VS Code even highlights it as an error:

image

_Why did CI not catch this? 🤔 _ (looks at `test`) Oh I guess I don't run TS on the `src` folder yet, just on `__tests__`, because not enough of the source code is actually typed.

Issue #714, opened 2019-03-21
Fetching single commit


hirokiosame:
Looking to be able to download a single specific commit.

Looks like git fetch is able to do that:
[`git fetch --depth=1 ../testrepo/.git $SHA1`](https://stackoverflow.com/questions/14872486/retrieve-specific-commit-from-a-remote-git-repository/30701724#30701724)

Is there a way to do this with isomorphic git?

wmhilton:
Yup. Depends on whether the remote supports it (via `allow-tip-sha1-in-want` or `allow-reachable-sha1-in-want`) but I just tried this and it works:

```bash
mkdir tmp
cd tmp
isogit init
isogit clone --depth=1 --url=https://github.com/isomorphic-git/isomorphic-git --ref=9aa9c2aca7fc6086e0ced9fa886442beba7bcf79
```

willstott101:
This breaks with `singleBranch: true` due to the way the `refs` map is built in the `_fetch` function for `updateRemoteRefs`. But without `singleBranch`, whilst the command completes, the commit isn't found when I use `readCommit`. It seems to essentially fallback to `fetch --all` without `singleBranch`.

PR #1225, opened 2020-09-07
fix(fetch): fetching a specific commit hash with `singleBranch: true`

willstott101:
Addresses my comment https://github.com/isomorphic-git/isomorphic-git/issues/714#issuecomment-688215101

I guess there is no ref information to update on download. Not 100% sure about this fix, but the test definitely catches it.

Would it be better to skip calling `updateRemoteRefs` rather than calling with an empty Map?

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -434,6 +434,16 @@
       "contributions": [
         "bug"
       ]
+    },
+    {
+      "login": "willstott101",
+      "name": "Will Stott",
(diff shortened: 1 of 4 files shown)


wmhilton:
> Would it be better to skip calling `updateRemoteRefs` rather than calling with an empty Map?

Nah, I like to have as few conditional branching paths as possible so that works fine.

> This breaks with `singleBranch: true`

Good find! Thank you! 👍

Issue #1182, opened 2020-07-06
Reconsider git.add supporting multiple files


angusmunro:
Hi there!

Firstly let me thank you for this project - It has been very useful to us in implementing the git functions we want to support!

That said I am now looking into some key performance issues and am finding the git.add to be a pain point for us.

I am wondering if you would reconsider implementing multi file support for git.add .... I am asking because it appears to be very slow to add lots of files in a loop for _specific git repositories_. 

On average it is taking 46ms for each call to `await git.add` and we are looking to want to add 50-100 files at a time. Even on faster repositories its averaging around 30ms

perhaps something like this?
`await git.add(file1.txt folder/file2.txt file3.txt file4.txt file5.txt)`

crutchcorn:
I could add in this functionality relatively trivially if we decided we want to move forward with a specific API @wmhilton. We could likely even do it in a non-breaking manner.

However, @angusmunro, is there a reason you wouldn't want to use a pattern like this and provide your own `gitAddMany`:

```javascript
const gitAddMany = async (files) => {
    return Promise.all(files.map(file => git.add(file)))
}
```

angusmunro:
@crutchcorn Thank you for your reply!

Your snippet is a sound idea in the case where you are not wanting to commit the files straight after. However, if you have the workflow we are currently trying to optimise this becomes a problem.

I am trying to decrease the turnaround between the following:

-> Adding files -> Committing files -> Pushing the branch -> Informing the user it is done.

Perhaps it is ignorance on my part, but it seems that by adding the files individually we are experiencing this being slow when there are many files being committed. My hope is that by requesting a git.add functionality that supports multiple files that we would see a shorter overall time for this workflow; anecdotally I have seen doing a blanket git.add for lots of files not noticeably slower than staging a single file when I am working with git directly, rather that through isomorphic-git.

If you do not believe this would be the case then by all means let me know and I apologise for my ignorance!

Thanks again for replying! We are very appreciative for this npm package - It its awesome and the documentation has helped us no end in getting to where we are at today.

Kind regards,

Angus

wmhilton:
Hi! Sorry for delayed response but yes, absolutely we can and probably should do that.

> I am asking because it appears to be very slow to add lots of files in a loop for specific git repositories.

yes, there's some overhead reading and parsing git-related files (the .git/config file, the .git/index file, etc) so that doing many calls in a loop is (currently) slower than having a dedicated method to do it.

[...]

PR #1228, opened 2020-09-08
feature: Initial work to add many files using git add

crutchcorn:

This is the initial PR for the logic of #1182. However, I will note that the change made to errors thrown will be considered a breaking change for error handling. Do we want to change that so that it's no longer a breaking change?

### DIFF
--- a/__tests__/test-add.js
+++ b/__tests__/test-add.js
@@ -25,6 +25,14 @@ describe('add', () => {
     await add({ fs, dir, filepath: 'b.txt' })
     expect((await listFiles({ fs, dir })).length).toEqual(3)
   })
+  it('multiple files', async () => {
+    // Setup
+    const { fs, dir } = await makeFixture('test-add')
+    // Test
+    await init({ fs, dir })
+    await add({ fs, dir, filepaths: ['a.txt', 'a-copy.txt', 'b.txt'] })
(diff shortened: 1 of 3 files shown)


crutchcorn:
Going through my org's repos and wanting to delete our fork of `isomorphic-git`, as we're not relying on our fork anymore.

@wmhilton how are we feeling about this PR? I know there's a merge conflict here - if I clean that up do we feel comfortable to merge?

jcubic:
Sorry for the late reply but the project was in a suspended state, and I just become a new maintainer and I'm handling all issues, one by one.

I've discussed with @wmhilton and it would be better to have an overloaded parameter that accepts `string | string[]`. Can you make this change? Also, can you resolve the conflicts? The test failed I'm not sure if because of broken tests or because of initial conflict.

crutchcorn:
@jcubic no worries at all. I've worn the new maintainer hat a fair number of times myself. 

What do we want to call this overloaded prop? If we're going for backcompat, we'd obv do `filepath` altho ofc that's no longer true when using an array. I'm fine with that, just confirming that's what we're going with 

Tests and merge conflicts are ofc on the radar as well

jcubic:
yes, I've discussed with @wmhilton and it would much better to have a single `filepath`, actually, I don't mind this it's a single form. 

It's better to have single parameter instead of checking all combinations of two parameters.

PR #1232, opened 2020-09-12
Don't mark type: module in package.json until isogit doesn't use `require()` statements

### LINKED ISSUE #1224: ReferenceError: require is not defined
Using node 14 (I believe because of the `"type": "module"` in `package.json`) the CLI doesn't work.

The CLI uses `require` which is not available in an ESM context.

```
wstott@wstott-1804:/tmp$ mktemp -d tmpnode-XXXXXXXX
wstott@wstott-1804:/tmp$ cd tmpnode-YNkwT6GT/
wstott@wstott-1804:/tmp/tmpnode-YNkwT6GT$ ls
wstott@wstott-1804:/tmp/tmpnode-YNkwT6GT$ node --version
v14.8.0
wstott@wstott-1804:/tmp/tmpnode-YNkwT6GT$ npm init
wstott@wstott-1804:/tmp/tmpnode-YNkwT6GT$ npm i isomorphic-git
wstott@wstott-1804:/tmp/tmpnode-YNkwT6GT$ npx isogit
file:///tmp/tmpnode-YNkwT6GT/node_modules/isomorphic-git/cli.js:2
const fs = require('fs')
           ^

ReferenceError: require is not defined
    at file:///tmp/tmpnode-YNkwT6GT/node_modules/isomorphic-git/cli.js:2:12
    at ModuleJob.run (internal/modules/esm/module_job.js:140:23)
    at async Loader.import (internal/modules/esm/loader.js:165:24)
    at async Object.loadESM (internal/process/esm_loader.js:68:5)
```

If I edit the `package.json` in `node_modules` to not include `"type": "module",` then the CLI starts working.

snickell:
As of node v14 (I believe, maybe earlier?), `"type": module` indicates that JS files should be interpreted as ".mjs", which means that using `import` statements instead of `require` is mandatory. This makes `npx isomorphic-git` etc not work out of the box, since `cli.js` still uses require().

The alternative fix to this (which may require changing a lot of imports in tests too?) would be to replace all require() calls with import statements, but I believe this fix is less likely to work across a larger span of node versions for the time being.

After applying this PR, `node cli.js` works in node v14.

Fixes #1224


snickell:
BTW, I'm happy to "squash merge the PR" but I don't know how, I've always done the squashing via github when merging the PR.... happy to push a squashed version if I know what to do :-)

wmhilton:
> `npx isomorphic-git`

People actually use the CLI? 👀 Hmm. Hm. Let me see if I can make it work by just sticking the CLI in it's own directory with its own `package.json`. I'd rather not give up my coveted `"type": "module"` since I did a ton of work to get that working properly in Node 13. (Obviously not for the CLI, but for the JS API I made it work.)

> BTW, I'm happy to "squash merge the PR" but I don't know how, I've always done the squashing via github when merging the PR.... happy to push a squashed version if I know what to do :-)

Oh yeah that last step is for me. I guess I should modify the PR template.

wmhilton:
Or maybe I can just rename `cli.js` to `cli.cjs`? 🤔

wmhilton:
Yes, I believe I can! (just rename `cli.js` to `cli.cjs`)

PR #1247, opened 2020-10-15
fix: remove fallthrough switch statement

mmkal:

This removes the use of a fall-through switch statement. While it technically works, it's confusing to humans - and, as it turns out, to machines.

I have a slightly strange use case, where the [runtime I'm using](https://github.com/mmkal/plv8-git) can't handle async/await. I'm able to make isomorphic-git work by supplying it with a filesystem with purely sync operations. But to avoid promises, I'm using [babel-plugin-transform-async-to-promises](https://npmjs.com/babel-plugin-transform-async-to-promises), which converts async/await statements into `.then`/`.catch` wrappers.

This allows shimming the global `Promise` object to do everything synchronously (or throw, in cases like http calls where that's impossible). Unfortunately babel-plugin-transform-async-to-promises can't handle switches with fallthrough. So most likely this is technically a bug with that package rather than isomorphic-git, but fixing it there would be **really** complicated since it involves dynamic code-generation, and requires very careful thought about edge-cases. Plus, the repo isn't very active. Fixing it here makes the code clearer, IMO and avoids the violation of a lint rule.

I haven't added or changed any tests because the functionality should be identical. If the existing tests pass, it should be safe to merge.

cc @wmhilton


src/storage/readObject.js
```diff
@@ -55,36 +55,31 @@ export async function _readObject({
...
-        )
-      }
-      const { object, type } = GitObject.unwrap(result.object)
-      result.type = type
-      result.object = object
-      result.format = 'content'
-    }
-    case 'content': {
-      if (format === 'content') return result
-      break
```
mmkal, line 83: note: I don't think this `break` statement should have been here. I've kept the equivalent for consistency, by converting to a `return`, but by breaking here, we'd end up returning undefined instead of throwing, which seems wrong. I'd be happy to open another PR fixing that issue as a follow-up - I'd rather keep the functionality identical with this PR, in case there are side effects to the fix.

mmkal:
@wmhilton any thoughts on this?

mmkal:
@jcubic conflicts are now resolved, this is ready to merge!

mojavelinux:
I've looked at this code more than a few times and always thought to myself it needed to be changed. So I'm in support.

FYI, William isn't currently working on the project, so it is best not to mention him. The current maintainers already get notifications when a PR is submitted, so there's really no need to use the @ mention unless you are replying directly to that person.

wmhilton:
```js
// BEHOLD! THE ONLY TIME I'VE EVER WANTED TO USE A CASE STATEMENT WITH FOLLOWTHROUGH!
```

Hahaha. So in the end, even this only time was a mistake! 😂

Good job y'all!

PR #1270, opened 2020-11-25
Two commits from master missing in main

chmac:
I guess there were 2 PRs opened before master was renamed to main, and when those two PRs were merged, they landed in the wrong branch.

One of these 2 commits is a fix for react-native which is not in the currently released 1.8.0 version. I've just hit the problem, and by pure random fluke, found the fix. I'd be delighted to see 1.8.1 shipped with that fix included! :-)


wmhilton:
man I really botched the master -> main transition. :(

PR #1280, opened 2020-12-08
resolves #1279: avoid traversing index when tree provided in _commit

### LINKED ISSUE #1279: Commit: Don't look for index when tree provided
There doesn't appear to be any reason to find the index and traverse it when a tree is provided to commands/commit (_commit).

djencks:
## I'm fixing a bug or typo

code seems a bit confused, I moved the index traversal so it's only done when needed.

### DIFF
--- a/src/commands/commit.js
+++ b/src/commands/commit.js
@@ -59,54 +59,54 @@ export async function _commit({
     })
   }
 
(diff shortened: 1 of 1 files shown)


jcubic:
Unfortunately, I cannon merge your PR because it broke tests.

PR #1316, opened 2021-02-18
fix: GitIgnoreManager to use exclusion rules from .git/info/exclude

nelsonni:
Adding support for exclusion rules from the `$GIT_DIR/info/exclude` file (per https://git-scm.com/docs/gitignore), based on the `TODO` tag in `GitIgnoreManager.js`:

https://github.com/isomorphic-git/isomorphic-git/blob/89c0da78d5ebf3c9f2754b3c8d557155dd70c8d7/src/managers/GitIgnoreManager.js#L7-L12

The contents of `$GIT_DIR/info/exclude` are in the same format as any `.gitignore` file, so reading and parsing the rules should be the same. Since the ignore manager from [`ignore`](https://www.npmjs.com/package/ignore) already handles removing comments and any other non-functional lines, we can simply add both rulesets into the ignore manager via [`ignore.add(patterns: Array<string | Ignore>)`](https://www.npmjs.com/package/ignore#addpatterns-arraystring--ignore-this).


src/managers/GitIgnoreManager.js
```diff
@@ -6,15 +6,14 @@ import { join } from '../utils/join.js'
...
-// TODO: Implement .git/info/exclude
-
 export class GitIgnoreManager {
   static async isIgnored({ fs, dir, gitdir = join(dir, '.git'), filepath }) {
     // ALWAYS ignore ".git" folders.
     if (basename(filepath) === '.git') return true
     // '.' is not a valid gitignore entry, so '.' is never ignored
     if (filepath === '.') return false
+    // Load exclusion rules from the project exclude file (.git/info/exclude)
+    const excludes = await fs.read(join(gitdir, 'info', 'exclude'), 'utf8')
```
jcubic, line 16: I would add check if the file exists first.

jcubic (changes requested):
Thanks for the PR.

src/managers/GitIgnoreManager.js
nelsonni: Fixed in dcad82d.

wmhilton:
This is fantastic!!! Thanks!!

PR #1382, opened 2021-06-30
posixify symlink path when adding file to repository

### LINKED ISSUE #1381: add command does not write posix path for symlink on Windows
Symlinks in a git repository must be expressed using a posix path (the path separator should be a forward slash) (e.g., `dir/file`). On Windows, the add command in isomorphic-git writes the path using backslash as the path separator. This happens because isomorphic-git uses fs.readlink to read the link target. On Windows, this function returns the path using backslash as the path separator (e.g., `dir\file`). isomorphic-git does not correct this, and ends up writing the non-posix path into the blob object.

This can be fixed by checking for the existence of backslashs in the return value of fs.readlink and correcting them to forward slashes.

Platform: Node.js
isomorphic-git version: 1.8.6

mojavelinux:
If backslashes are detected in return value of fs.readlink, change them to forward slashes. This logic is handled by the new posixifyPathBuffer helper, which filters the return value of fs.readlink.

I've also added a test to verify that symlinks can be added to the repository and that the contents of the symlink is the target path in posix form.

closes #1381

### DIFF
--- a/src/utils/posixifyPathBuffer.js
+++ b/src/utils/posixifyPathBuffer.js
@@ -1,0 +1,5 @@
+export function posixifyPathBuffer(buffer) {
+  let idx
+  while (~(idx = buffer.indexOf(92))) buffer[idx] = 47
+  return buffer
+}
--- a/__tests__/test-add.js
+++ b/__tests__/test-add.js
@@ -1,5 +1,12 @@
 /* eslint-env node, browser, jasmine */
-const { init, add, listFiles } = require('isomorphic-git')
+const {
+  init,
+  add,
+  listFiles,
+  readBlob,
+  walk,
+  STAGE,
+} = require('isomorphic-git')
 
 const { makeFixture } = require('./__helpers__/FixtureFS.js')
 
@@ -9,6 +16,10 @@ const writeGitIgnore = async (fs, dir) =>
     dir + '/.gitignore',
     ['*-pattern.js', 'i.txt', 'js_modules', '.DS_Store'].join('\n')
   )
+
+// NOTE: we cannot actually commit a real symlink in fixtures because it relies on core.symlinks being enabled
+const writeSymlink = async (fs, dir) =>
+  fs._symlink('c/e.txt', dir + '/e-link.txt')
 
 describe('add', () => {
   it('file', async () => {
@@ -24,6 +35,36 @@ describe('add', () => {
     expect((await listFiles({ fs, dir })).length).toEqual(2)
     await add({ fs, dir, filepath: 'b.txt' })
     expect((await listFiles({ fs, dir })).length).toEqual(3)
+  })
+  it('symlink', async () => {
+    // Setup
+    const { fs, dir } = await makeFixture('test-add')
+    // it's not currently possible to tests symlinks in the browser since there's no way to create them
+    const symlinkCreated = await writeSymlink(fs, dir)
+      .then(() => true)
+      .catch(() => false)
+    // Test
+    await init({ fs, dir })
+    await add({ fs, dir, filepath: 'c/e.txt' })
+    expect((await listFiles({ fs, dir })).length).toEqual(1)
+    if (!symlinkCreated) return
+    await add({ fs, dir, filepath: 'e-link.txt' })
+    expect((await listFiles({ fs, dir })).length).toEqual(2)
+    const walkResult = await walk({
+      fs,
+      dir,
+      trees: [STAGE()],
+      map: async (filepath, [stage]) =>
+        filepath === 'e-link.txt' && stage ? stage.oid() : undefined,
+    })
+    expect(walkResult.length).toEqual(1)
+    const oid = walkResult[0]
+    const { blob: symlinkTarget } = await readBlob({ fs, dir, oid })
+    let symlinkTargetStr = Buffer.from(symlinkTarget).toString('utf8')
+    if (symlinkTargetStr.startsWith('./')) {
+      symlinkTargetStr = symlinkTargetStr.substr(2)
+    }
+    expect(symlinkTargetStr).toEqual('c/e.txt')
   })
   it('ignored file', async () => {
     // Setup
--- a/src/api/add.js
+++ b/src/api/add.js
@@ -8,6 +8,7 @@ import { FileSystem } from '../models/FileSystem.js'
 import { _writeObject } from '../storage/writeObject.js'
 import { assertParameter } from '../utils/assertParameter.js'
 import { join } from '../utils/join.js'
+import { posixifyPathBuffer } from '../utils/posixifyPathBuffer.js'
 
 /**
  * Add a file to the git index (aka staging area)
@@ -69,7 +70,7 @@ async function addToIndex({ dir, gitdir, fs, filepath, index }) {
     await Promise.all(promises)
   } else {
     const object = stats.isSymbolicLink()
-      ? await fs.readlink(join(dir, filepath))
+      ? await fs.readlink(join(dir, filepath)).then(posixifyPathBuffer)
       : await fs.read(join(dir, filepath))
     if (object === null) throw new NotFoundError(filepath)
     const oid = await _writeObject({ fs, gitdir, type: 'blob', object })
--- a/src/models/FileSystem.js
+++ b/src/models/FileSystem.js
@@ -201,7 +201,8 @@ export class FileSystem {
     // Note: FileSystem.readlink returns a buffer by default
     // so we can dump it into GitObject.write just like any other file.
     try {
-      return this._readlink(filename, opts)
+      const link = await this._readlink(filename, opts)
+      return Buffer.isBuffer(link) ? link : Buffer.from(link)
     } catch (err) {
       if (err.code === 'ENOENT') {
         return null


jcubic:
There are broken unit tests:
```
[Mobile Safari 13.0.0 (iOS 13.0.0) | add | symlink]: Expected 'Hola!' to equal 'c/e.txt'.
[Chrome 79.0.3945 (Windows 10 0.0.0) | add | symlink]: Error: Expected 'Hola!' to equal 'c/e.txt'.
[HeadlessChrome 0.0.0 (Linux 0.0.0) | add | symlink]: Error: Expected 'Hola!' to equal 'c/e.txt'.
```

mojavelinux:
Those failures are actually unrelated to the code change itself (since they are following a different branch). The assertion would fail either way. The problem seems to be that in those cases, the symlink is not set up correctly in the fixture, hence why it is resolving the contents of the file. I think I know what I need to do to address it. Stay tuned.

jcubic:
I was not saying that the code was wrong. The fix may involve updating the fixture/unit tests to match new symlink behavior. They probably just broken git symlinks anyway. Your change will correct that but you also need to update the fixtures, so tests pass.

mojavelinux:
Yes, I get that. We are saying the same thing. I was just explaining what was going on.

mojavelinux:
...still working on this. I'll let you know once I got it.

mojavelinux:
It doesn't look like it is possible to create a symlink in the browser environment. Therefore, we cannot test that isomoprhic-git adds the symlink correctly. So what I'm going to do is check if the symlink creation succeeds (such as in Node.js) and only test it if it does...which is the only thing we need to test anyway since this is fundamentally a Node.js issue.

mojavelinux:
@jcubic All clear!

mojavelinux:
I'm working on one more small change.

mojavelinux:
I was able to verify that the test runs in the browser by enabling LightningFS. In order to do that, I needed to pass the following line to DefinePlugin in karma.conf.js

```
'process.env.ENABLE_LIGHTNINGFS': `true`,
```

That makes me realize we aren't actually testing LightningFS in CI. If we were, we'd see that its readlink function returns a String instead of a Buffer. So if we want to support LightningFS, it's necessary to use the fallback I added to posixifyPathBuffer.

mojavelinux:
@jcubic I'm confident this is ready to go now.

src/utils/posixifyPathBuffer.js
```diff
@@ -0,0 +1,8 @@
+export function posixifyPathBuffer(buffer) {
+  if (buffer instanceof Buffer) {
+    let idx
+    while (~(idx = buffer.indexOf(92))) buffer[idx] = 47
+    return buffer
+  }
+  return Buffer.from(buffer.startsWith('./') ? buffer.substr(2) : buffer)
+}
```
jcubic, line 8: I think that the place where it's added it not optimal. I would just add this to [FileSystem proxy](https://github.com/isomorphic-git/isomorphic-git/blob/main/src/models/FileSystem.js). That is a facade for the `fs` client. You can there return Buffer like in [read method](https://github.com/isomorphic-git/isomorphic-git/blob/main/src/models/FileSystem.js#L66).

You should do all manipulation in FileSystem object.

Also, I have a question: why you don't use `Buffer.from(buffer.replace(/\\/g, '/'))` on a string the same as what you do with buffer?

mojavelinux: > why you don't use Buffer.from(buffer.replace(/\\/g, '/')) on a string the same as what you do with buffer?

When the type is a buffer, I manipulated it directly as a performance optimization.

mojavelinux: > I think that the place where it's added it not optimal. I would just add this to FileSystem proxy. That is a facade for the fs client. You can there return Buffer like in read method.

In this case, I don't think that's the right choice. The reason is, it's not wrong that the fs.readlink returns a path with the system-dependent separator. But it is wrong for git.add. So we are modifying it at the exact boundary of where it should be modified. So I don't agree this should be done by the FileSystem object.

mojavelinux: (Usually, I am the one making the argument for using FileSystem object. But it's just not appropriate in this case).

What is wrong is that LightningFS should be returning a Buffer, not a string. And it should not be adding `./` to the start of that string. But those are bugs in LightningFS and we need to patch them until a new version of LightningFS is released. Then we can come back and remove the workaround.

strangedev: I agree with @mojavelinux that the `FileSystem` proxy should not contain this behavior. `FileSystem` should not have any knowledge about git IMHO. Adding the transformation the the proxy will impose git-specific behavior onto it, making it harder to re-use in other use-cases.

I've got a different concern with this particular piece of code though and that is that the behavior (removing `./`  __or__ removing backslashes) is controlled by the type of `buffer` that is passed into the function. I get that if it's a string, we're dealing with LightningFS and want to treat it differently. It just seems to me that the fact LightningFS uses a string instead of a buffer is coincidental and not clear to the reader. IMHO the code would be more readable if we'd first check the type, convert it to buffer if it's a string and then apply the replacements, regardless of which specific fs implementation is used.

It just doesn't seem right to me to detect the fs implementation based on this bug.

Maybe a solution would be adding the type conversion into the proxy, so that buffers are always returned by `fs.readlink()`. The logic for dealing with backslashes and prefixes would still be separate, but operate on buffers. Also, the replacement logic could be written in such a way that it can deal with all known issues and not assume the presence of certain bugs in the fs implementaion.

jcubic: > Maybe a solution would be adding the type conversion into the proxy, so that buffers are always returned by fs.readlink(). The logic for dealing with backslashes and prefixes would still be separate, but operate on buffers. Also, the replacement logic could be written in such a way that it can deal with all known issues and not assume the presence of certain bugs in the fs implementation.

This is exactly what I've written in the previous message but with different words.

> I think that returning Buffer from readlink should be done FileSystem object

And before patching I would check what exactly is the result of fs.readlink on WIndows and check if lightingFS return correct results, if the path is posix like then I would patch lighintgFS and if not then I will check how canonical git does this
on Windows, and do the same.

Also, Note that FileSystem object is only for git stuff, it's a wrapper for fs client. FileSystem is not suppose to be used bot anyone except git.

jcubic: But if the patch is only needed because of a bug in lightingFS I would fix lightingFS instead of introducing this temporary patch that will be removed anyway.

strangedev: > This is exactly what I've written in the previous message but with different words.

Sorry @jcubic, didn't see your message. Am a bit puzzled why the order of messages is messed up. Your message shows as "7 minutes ago" for me and my message as "1 hour ago". Full agree on the points you raised though.

jcubic: It seems that we just found a bug in GitHub. I've added one comment you added later and I was not able to comment to your comment and by mistake added my reply as a normal comment not in review and my comment that you've missed is in two places with same time.

jcubic: It seems that we just found a bug in GitHub. I've added one comment you added later and I was not able to comment to your comment and by mistake added my reply as a normal comment not in review and my comment that you've missed is in two places with same time.

jcubic: It seems that we just found a bug in GitHub. My comment that you've missed is in two places at the same time.

mojavelinux: > it return 3 completely different not compatible results:

I'm not trying to make a generic function. I'm trying to make a function that gets the job done it needs to do. And it does that, as the tests prove.

mojavelinux: I think we are bikeshedding the heck out of this problem. The code is simple and the tests pass. What's the motivation for blocking this? We need to fix the problem in isomorphic-git.

Okay, so LightningJS has a bug. If that's so much of a problem, let's just remove the workaround for it and say you can't use LightningJS until the bug in it is fixed. Frankly, I don't care about. I was just trying to be a good citizen of isomorphic-git. What I care about is Node.js.

mojavelinux: > Why not add path fix to lightingFS, it will make much more sense.

Because that is outside the scope of what I'm willing to commit time on right now. I've already spend 2 whole days on this patch...just to switch backslashes to forward slashes.

strangedev: I'm sorry this turned out so frustrating your you @mojavelinux. FWIW, I just saw that this PR is open for a couple of days and wanted to give some input. 

I'd also like to see this fixed and frankly, I don't think adding workarounds for 3rd party packages is in the scope of this project.

What irks me is that the implementation provided only strips out backslashes when a buffer is passed. This doesn't seems architecturally sound to me, but I also concede this might be the simplest way to fix this at the moment. 

My worry is that this function will become another difficult to understand part in an already somewhat confusing codebase. I say at least add a comment why this works.

mojavelinux: > What irks me is that the implementation provided only strips out backslashes when a buffer is passed.

It's only ever supposed to be a Buffer. I could add a comment as to why there is a fallback with a note that it should be removed when LightningFS is fixed...or, like I suggested, we can just get rid of it and not worry about LightningFS.

strangedev: > It's only ever supposed to be a Buffer.

Right. And it's also not supposed to start with `./`, correct? I'd be okay with handling backslashes as the default case and handling anything else as a fallback with a note describing the case (the type being string, the prefix and that it can't contain backslashes by nature).

These are just my two cents though, @jcubic is calling the shots.

strangedev: I don't know how popular LightningFS is, but removing it seems a little radical to me.

mojavelinux: When it is not a Buffer, we know 100% that it's coming from LightningFS since it's the only fs not conforming to the spec by honoring the encoding option. (See https://github.com/isomorphic-git/isomorphic-git/blob/04fa3d9922783341ce2e4c5a45c4a9b3db69dd17/src/models/FileSystem.js#L201). That's where we could handle the translation to a Buffer (and perhaps even deal with removing the invalid leading `./`).

I'm willing to make a change, but I need some clear direction.

mojavelinux: > I don't know how popular LightningFS is, but removing support for it seems a little radical to me.

I'm not saying removing LightningFS outright. I'm saying removing the workaround that would enable it to pass the assertion in the test. (As it turns out, the leading `./` is benign if the value is passed to path.join because it will get cleaned up then...but it breaks our assertion). I could even update the assertion to allow it to be there...and thus just remove this whole workaround.

strangedev: > That's where we could handle the translation to a Buffer (and perhaps even deal with removing the invalid leading ./).

I think this is a good place to deal with the LightningFS specific things.

Removing the backslashes could be done in your helper still. I think this is also what @jcubic suggested before.

strangedev: > I'm saying removing the workaround that would enable it to pass the assertion in the test.

I'd like to ultimately see the workaround removed from the project. As it stands right now, it is probably more beneficial to have it, so that we don't break any existing integrations. We should probably keep an eye out on LightningFS and see if this gets fixed. 

To reiterate, I think the proxy is a good place for this workaround as it separates the fix you wanted to contribute originally from LightningFS weirdness.

mojavelinux: I updated the code. See what you think now.

I am still coercing the return value of readlink to a Buffer, but doing that part in FileSystem. That way, it adapts to the incorrect return value by LightningFS is a way that is very benign and forgiving. We can remove this eventually, but for now it allows symlinks to work when using LightningFS (which would otherwise crash on byteLength property not found due to the incorrect type).

mojavelinux: Can I use `.then` in this case?

```js
return this._readlink(filename, opts).then(link => Buffer.isBuffer(link) ? link : Buffer.from(link))
```

I ask since `then` is not used anywhere else in that file...so I'm fine with just keeping it as it is.

strangedev: I like it 👍

Thanks for not bailing on this rather lengthy discussion 😅

I believe this is small change improves comprehensibility a lot and while it took us some time to arrive at a consensus, I think it was worthwhile. 

Code is read many more times than it is being read and it's important we get this right and improve maintainability moving forward. 

Let's see what @jcubic thinks 🤔

strangedev: > Can I use .then in this case?

Seems to me that all of these internal methods are promisified, so yes, you should be able to use `.then` here.

mojavelinux: Thanks for working to get to that consensus.

>  so yes, you should be able to use .then here.

I was asking about code style. Is it acceptable or do we prefer the use of await with a local const?

mojavelinux:
CI is green again. Let me know what you think.

src/utils/posixifyPathBuffer.js
jcubic: This looks great. BTW: You've said in the previous comment that it was ok because tests were passing, but it was the problem with tests that they didn't check all the cases that I've shown in the comment. Now that the code is always executed I think it's fine.

But what about starting `./` is that not the problem if it's included? In the first code, you remove that from the string. Does it need to be removed in buffer?

mojavelinux: After looking into more, I was wrong that the extra leading `./` breaks the usage of isomorphic-git, even though it is clearly different than what is returned in other environments. Therefore, I updated the the assertion to allow for it.

mojavelinux:
Thanks again for working through this with me!

PR #1386, opened 2021-07-02
Chore  update deps

strangedev:
Heya :wave: 
I noticed that there are many dependabot PRs. Rather than trying to merge them all separately, I ran `npm-check` to update all non-major dependencies. There are still a lot of outdated deps, but this should reduce the burden somewhat. There were some issues with the way `ArrayBuffer` was compared using karma, so I updated karma (they added support for comparing `Buffers` somewhere along the way).


jcubic:
Thanks I was running `npm audit --fix`, to update any dependencies. But I should just update all dependencies.

I think that I will just close all dependabox PRs.

strangedev:
Yes I think closing the PRs and performing the upgrades manually is the way to go. We can start using dependabot once we reach a good and up-to-date state.

jcubic:
There are also a lot of vulnerabilities detected by GitHub they are all for devDependencies, it would be nice to update everything to the latest major versions. But we will need to fix the build scripts first. The good news is that it works on CI, so this don't stop us with releasing new versions.

strangedev:
Seems like Saucelabs is not working. I guess these tests are not executed locally. Unfortunately I don't have time to check it out at the moment, but I suspect the puppeteer version might have something to do with it. Will check it out later.

DecliningLotus:
I tend to use [Depfu](https://depfu.com/) for its ability to "group" updates. It checks dependencies on an interval of your choosing and sends a single PR for all the updates. The only individual PRs you would see then are for major updates. Supports monorepos too, so you would get different grouped updates for the main source and the website package.json.

Just a suggestion which I think will hugely remove a lot of the noise in this repo and makes things far more manageable.

An [example](https://github.com/fontsource/fontsource/pull/284) to give you guys an idea of what each PR looks like.

jcubic:
@strangedev can you make some commit to the PR? I'm not able to rerun the test. Maybe remove the change from unit tests and commit.

PR #1387, opened 2021-07-03
 use built-in fs.rm if available; otherwise, use fallback utility

### LINKED ISSUE #1383: Improve recursive rm
This is a continuation of the discussion in #1283.

The idea is to use fs.rm if it's present that allow for recursive delete, the docs will be updated, so it will be required to implement in new code. fs.rm is also available in the newest Node.

mojavelinux:
- add rm utility as fallback if fs does not provide rm
- delegate to fs.rm if recursive option is passed to FileSystem#rmdir
- remove deleteRecursively utility

resolves #1383


src/utils/rm.js
```diff
@@ -0,0 +1,20 @@
...
+  const entries = await fs.readdir(filepath)
+  if (entries == null) {
+    await fs.rm(filepath)
+  } else if (entries.length && opts.recursive) {
+    await Promise.all(
+      entries.map(entry => {
+        const subpath = join(filepath, entry)
+        return fs.lstat(subpath).then(stat => {
+          if (!stat) return
+          return stat.isDirectory() ? rm(fs, subpath, opts) : fs.rm(subpath)
```
jcubic, line 13: What about using `fs.unlink` so we are sure that we don't use any logic in FileSystem rm and end up with a recursive call, it may work but it's hard to unwrap this in head. And It will also work with symbolic links.

src/utils/rm.js
```diff
@@ -0,0 +1,20 @@
+import { join } from './join'
+
+export async function rm(fs, filepath, opts = {}) {
+  const entries = await fs.readdir(filepath)
+  if (entries == null) {
+    await fs.rm(filepath)
```
jcubic, line 6: I would use here a utility that will check if filepath is directory or not like the other commented line (and do fs.rmdir or fs.unlink), and use it here. So it's clear what is calling what.

jcubic (changes requested):

src/utils/rm.js
jcubic: Another idea is to use hidden `_rm` so it's clear that it's not a recursive call, because it looks weird that you in one place call `rm()` and in other `fs.rm()`.

mojavelinux:
I think it's best if we *do* use the FileSystem model because that gives us a lot of existing functionality we can leverage. We are trying to keep this code down in size so we don't have duplicate code here, specifically error handling logic. I much prefer how it is written now.

src/utils/rm.js
mojavelinux: It may look confusing because the functions are similarly named, but they are distinctively not the same thing. `fs.rm` is effectively `unlink` with a graceful exit if the file is already gone.

jcubic: Then maybe just add a comment that `fs.rm` is basically `unlink`.

mojavelinux: Sure thing. I'd also be open to renaming the rm helper function since we're only ever using it to remove recursively. I think rmRecursive (or rmRecursively) makes sense.

I'll update accordingly.

src/utils/rm.js
mojavelinux: I'm pretty sure that the code already supports removing symlinks. But regardless, that's outside the scope of this issue. The goal of this issue to make recursive rm more robust (for a gitdir that isomorphic-git itself created). I think we should stay focused on that goal.

mojavelinux:
I believe the concerns have been addressed. Let me know if you agree.

If you feel very strongly about it, I'm willing to rewrite the code so that it uses the underscore methods and does its own handling of files that have gone missing. Please advise if that's what you want me to do. As I said, I prefer reusing existing logic, but I don't want that preference to cause this to get blocked.

jcubic (approved):

jcubic:
I think that this is ok. Thanks for working on this.

jcubic:
Are there any changes that require the release of a new version?

mojavelinux:
Since we provide a fallback, I don't think a new minor version is warranted. This is really all happening under the covers.

mojavelinux:
Can we move forward with this?

jcubic:
Sorry merged, It seems that on my new system I didn't get an email. I need to figure out why the email went into the trash directory. The rules are the same as in my previous system. I have trash for evident SPAM that doesn't need filtering by the spam filter.

jcubic:
Ok, I've figured that out I've had a filter that was throwing emails with "chore" into the trash since there were a lot of them.
And I use rule that prevents throwing Azure emails into the trash but it was a rule `or not azure`. That rule was applied to every email I've got, because or not azure was in fact every email.

mojavelinux:
Thanks @jcubic!

PR #1405, opened 2021-08-02
docs: add faster approach to staging all files to faq

### LINKED ISSUE #1402: Safe to recommend `git.add({ dir, filename: '.' })` in FAQ entry for staging all untracked files?
Hi folks! I'm looking into opportunities to optimize the performance of a system which uses isomorphic-git. It needs to stage every file in a new repository in order to create an initial commit containing all the files. It currently does this by using globby to find all the files, then adding each file in one by one, following [the recommended approach in the docs](https://isomorphic-git.org/docs/en/faq#how-to-add-all-untracked-files-with-gitadd). I've found that this approach takes minutes to complete when the number of files is large. Conversely, passing the top level directory of the repo using `git.add({ dir: '<repo root>', filename: '.' })` seems to work correctly for the use cases I've tested, and completes in a comparable time to running `git add .` in the terminal.

With this in mind, I'm wondering whether this is a safe change for us to make. If so, I'd be happy to send a PR updating the recommended approach for this use case in the FAQ!

mtlewis:
Closes #1402.

Further to the discussions in the above issue, I'm opening a PR to update the FAQ with a different recommendation for adding all files in a repository using `git.add`. I did some testing to measure the comparative performance of the two approaches. Having created a bare repo with `git init`, I created 10,000 small text files in a subdirectory (filename set to <uuid>.txt, content set to <uuid>), and ran the old and new versions of the answer against it (unstaging between runs). The old approach took just over 9 minutes to add all the files, while the new version took just under 9 seconds.

I'm still really keen to get some feedback on whether this makes sense - @wmhilton, wondering if you might be able to shed some light on whether there are pitfalls with the proposed approach that means it's not appropriate to recommend? If there are pitfalls, it seems like given how much faster `git.add({ fs, dir, filepath: '.' })` is, it might still be worth mentioning as an option for people who understand and are willing to accept the drawbacks.


jcubic:
You can add yourself as a contributor that added docs and also as the author of the solution in FAQ.

mtlewis:
@jcubic done!

jcubic:
Thanks.

PR #1427, opened 2021-10-09
Make GitConfig parser more resilience

eugeniol:
It seems git.clone method only writes the config if [corsProxy is enabled](https://github.com/isomorphic-git/isomorphic-git/blob/main/src/commands/clone.js#L68), this causes all other methods that rely on getConfig to fail since the config does not exist.

### DIFF
--- a/src/models/GitConfig.js
+++ b/src/models/GitConfig.js
@@ -124,7 +124,7 @@ export class GitConfig {
   constructor(text) {
     let section = null
     let subsection = null
-    this.parsedConfig = text.split('\n').map(line => {
+    this.parsedConfig = (text || '').split('\n').map(line => {
       let name = null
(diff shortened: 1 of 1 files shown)


jcubic:
Thanks for the PR I need to look closer, because it seems that you just fix side-effects not the root cause. I need to check where and how this config is used.

jcubic:
Sorry, for the late reply I will see if I will find time on a weekend and read the code to see where this config is used and it's ok to just use what you propose.

I need to be sure that there are no bugs in a different place where it passes null but it should not.

It's silly but only now I've found how to show all PR that doesn't come from dependabot (`-label:dependabot`).

Additional thing if you still want this fix to be merged, it would be nice to see at least one test using public API.

PR #1448, opened 2021-12-30
Fixes in collecting uint8array unexpected buffers

rlaurente:
src/utils/collect.js 
- throws error when pushing updates.
- Error comes in in http/web collecting uint8arrays has unexpected buffer item

### DIFF
--- a/src/utils/collect.js
+++ b/src/utils/collect.js
@@ -5,9 +5,19 @@ export async function collect(iterable) {
   const buffers = []
   // This will be easier once `for await ... of` loops are available.
   await forAwait(iterable, value => {
-    buffers.push(value)
-    size += value.byteLength
-  })
+
+    //  catch some unexpected buffer in array of uint8arrays
+    if(!value.byteLength){
+      let ui8 = new Uint8Array(value.length);
+      for (var i = 0; i < value.length; i++) {
+        ui8[i] = value[i];
+      }
+      value = ui8;
+    }
+
+    buffers.push(value);
+    size += value.byteLength;
+  });
   const result = new Uint8Array(size)
   let nextIndex = 0
   for (const buffer of buffers) {


mojavelinux:
> Error comes in in http/web collecting uint8arrays has unexpected buffer item

It seems to me like the error should be addressed there. I don't like the unnecessary defensive programming. The value is assumed to respond to `byteLength`, and thus it should if it is going to use this function.

rlaurente:
Hi @mojavelinux 

I just moved the code to http/web and change the condition to make it more understandable. Please let me know if that works.

Thanks

mojavelinux:
I still don't understand the use case. In what situation is this function even coming across such a value? I would like to see a test for this that proves when it happens. Are we guarding against a non-compliant HTTP client? Please explain.

rlaurente:
I did basic clone, add changes, commit and push in browser and I came across this error.

Screen-Shot-2021-12-29-at-8 04 43-PM

This is the basic sample code below.

const commit_result = await git.commit({
        ...repo,
        author: {
            name: 'Tester',
            email: 'tester@gmail.com',
        },
        message: 'Nobe added updates'
    });
    console.log(commit_result, 'commit_result')

    const push_result = await git.push({
        ...repo,
        remoteRef: Config.BRANCH,
        remote: 'origin'
    })
    console.log(push_result);

I further debug that issue, see that last item of the array is buffer and that triggers that error above. So what I did detect any buffer and convert it to Uin8array.

Thanks

rlaurente:
Hi @mojavelinux 

Were you able to replicate the same issue above ? if not, I was thinking to make a codepen.

Thanks

mojavelinux:
There's still not enough usable information here for me to be able to reproduce what you are doing. Sharing a CodePen would be helpful. Even better would be to share a failing test in the test suite. I'm not going to give my approval to any change that doesn't have a test case that verifies the scenario and the solution. We can't just change code in this code base without backing it up with assertions.

PR #1453, opened 2022-01-10
fix(is-ignored): allow paths ending with /

SamVerschueren:
When using the `isIgnored` API, I noticed that providing `dir/` as `filepath` argument didn't work and throws an error `path must not be empty`. The reason is that the `GitIgnoreManager` iterates over the segments of the path, which in one case is an empty string.

Added a check for that and also added tests for this API. Apparently it wasn't tested yet.

## I'm fixing a bug or typo


jcubic:
Please fix lint errors, there is no reason to wrap the expression in extra parentheses. It's only needed if you call a method on return value but here you have

```javascript
expect(await x).toBe()
```

and await is in double parentheses.

SamVerschueren:
Fixed the linting issues but the tests are still failing. Looking into it.

jcubic:
What about if you just add `.filter(Boolean)` to exclude empty string as part of the path?

SamVerschueren:
@jcubic Yes you are correct! That feels a bit cleaner :).

I also found a bug in the `ignore` package that `isomorphic-git` uses. You can read more about it [here](https://github.com/kaelzhang/node-ignore/issues/77).

wmhilton:
Oh that's a wonderful enhancement! Thank you ❤️

wmhilton:
It makes me so happy to see folk contributing and @isomorphic-git-bot taking care of publishing... there's still things I wish were easier (running the tests and build still requires node 10 🤪 which is awfully out-of-date... something something experimental ECMAScript Modules stuff changed) but I tried to automate as much as I could. I never quite figured out how to make the repo auto-governed by an decentralized leaderless collective...

SamVerschueren:
Love the work you put in this project @wmhilton (and the rest of the folks hanging around here @jcubic). It's really cool to see a PR merged and be able to instantly use it! 💪

mojavelinux:
I second the sentiment shared by @SamVerschueren!

jcubic:
Yes, the CI/CD is great in this repo thanks to @wmhilton that build the thing, even if some parts of it are out of date.

PR #1461, opened 2022-01-23
fix: update the config parser to handle dots in key name

### LINKED ISSUE #1460: Config parser does not split on dots correctly
The config parser is naively splitting on all dot characters. This breaks if the key for a section contains a dot. Here's how git defines the parsing rules:

> the fully qualified variable name of the variable itself is the last dot-separated segment and the section name is everything before the last dot.

Let's assume the fully qualified variable name is as follows:

```
branch.v3.0.x.remote
```

The section should be "branch", the key "v3.0.x", and the variable name "remote".

This can be implemented using the following logic:

```js
const segments = path.split('.')
const section = segments.shift()
const name = segments.pop()
const key = segments.join('.')
```

mojavelinux:
fixes #1460

### DIFF
--- a/__tests__/test-GitConfig.js
+++ b/__tests__/test-GitConfig.js
@@ -276,6 +276,14 @@ describe('GitConfig', () => {
       url = https://foo.com/project.git`)
     })
 
+    it('existing subsection with dots in key', async () => {
+      const config = GitConfig.from(`[remote "foo.bar"]
+      url = https://foo.com/project.git`)
+      await config.set('remote.foo.bar.url', 'https://bar.com/project.git')
+      expect(config.toString()).toEqual(`[remote "foo.bar"]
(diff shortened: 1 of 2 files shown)


mojavelinux:
I'm not sure if this fix is comprehensive, but it does address the case when isomorphic-git uses config.set to set the remote tracking branch info.

mojavelinux:
We should probably add a test that it does create the right entry in .git/config when checking out a branch that contains a dot. So far, I only added unit tests for GitConfig itself.

jcubic:
Thanks for the fix.

jcubic:
Just noticed that you've said that there could be more tests. You can push new commits to this branch and create another PR if you want to add better tests.

mojavelinux:
See #1462 for additional test.

PR #1489, opened 2022-02-17
fix(resetIndex): allow resetting index in a new repository or with oid

SamVerschueren:
This PR is twofold. It fixes a bug, and introduces an additional option to `resetIndex()`. I'll describe both in the sections below.

## I'm fixing a bug or typo

- [x] squash merge the PR with commit message "fix(resetIndex): allow resetting index in a new repository"

When using [`resetIndex`](https://isomorphic-git.org/docs/en/resetIndex), I noticed that it does not work on an empty repository created with `git init`. Let's say you created a new repository and added some files to the index, resetting a file will not work because there are no commits yet, and thus `HEAD` will not resolve.

I will annotate the code and explain what and why. I also added an additional test which works on a new repository without commits. It did fail without the changes I made.

This is not the case anymore!

- [x] ~~add parameter to the function in `src/api/X.js` (and `src/commands/X.js` if necessary)~~
- [x] ~~document the parameter in the JSDoc comment above the function~~
- [x] ~~add a test case in `__tests__/test-X.js` if possible~~
- [ ] ~~if this is your first time contributing, run `npm run add-contributor` and follow the prompts to add yourself to the README~~
- [x] ~~squash merge the PR with commit message "feat(X): Added 'bar' parameter"~~

~~Currently [`resetIndex`](https://isomorphic-git.org/docs/en/resetIndex) only accepts a `ref` to reset the provided `filepath` to. This limits the possibilities, because `git reset` also accepts object ids pointing to a commit. I added an additional `oid` parameter which will get precedence over the `ref` option.~~

~~I added an additional test which resets the file based on a commit id.~~


src/api/resetIndex.js
```diff
@@ -26,38 +28,57 @@ import { resolveFilepath } from '../utils/resolveFilepath.js'
...
-    } catch (e) {
-      // This means we're resetting the file to a "deleted" state
-      oid = null
+
+    if (!oid) {
+      try {
+        // Resolve ref if no oid is provided
+        oid = await GitRefManager.resolve({ fs, gitdir, ref: ref || 'HEAD' })
+      } catch (e) {
+        if (ref) {
```
SamVerschueren, line 59: If a `ref` is explicitly provided, we rethrow the error indicating the ref could not be resolved. If no `ref` is provided, we just continue.

This is in analogy to how `git reset` works on an empty repository. If you run `git reset`, it will remove all the files from the index. If you run `git reset HEAD`, it will throw an error.

This is the reason I had to remove the default value for `ref` in the argument list.

src/api/resetIndex.js
```diff
@@ -26,38 +28,57 @@ import { resolveFilepath } from '../utils/resolveFilepath.js'
...
+          // Only throw the error if a ref is explicitly provided
+          throw e
+        }
+      }
+    }
+
+    if (oid) {
```
SamVerschueren, line 66: It only makes sense to resolve the file path of we have an `oid` at this point, if we don't have one we can skip it. The reason we don't have one is in the case of a `resetIndex()` without explicit `ref` on a new git repository.

src/api/resetIndex.js
jcubic: I would not remove the default `'HEAD'` from the parameter, this is almost the same as `ref || 'HEAD'`. But the default value in the function signature is some kind of documentation.

src/api/resetIndex.js
jcubic: Please add a comment with that info. It may be useful to have this in code.

SamVerschueren: Good idea! Added it.

jcubic:
I was wondering what if isomorphic-git handles all references in `ref` argument. Reference is a fundamental thing in GIT what about if we add a parser for references and it will handle all or most of them. Like `HEAD~5`, `<HASH>^2`, branch, or tag.

If we add `oid` as an argument we probably will not be able to add this later.
I'm thinking out loud.

SamVerschueren:
I like the idea for sure. I think everyone is customly implementing `HEAD~X` and what not themselves now. Having that baked into `isomorphic-git` makes a lot of sense. However, we might need to have a way to notify the user about ambiguous refs. If for instance the name of a branch is called `a2f3`, but we also have an object with that starts with that ID, git will warn the user. Although, maybe we don't want `ref` to be a short ID 🤔 .

I could see 2 paths forward here.

1. We just add `oid` and add this ref parsing later. Which would mean that when that lands, you should be able to do `resetIndex({ref: '<HASH>'})` instead of `resetIndex({oid: '<HASH>'})`. Currently `oid` always gets presedence over `ref`, so changing what `ref` accepts later should still be doable.
2. I change this PR so `ref` now also accepts `<HASH>`, but nothing else for now. It would not be entirely in line with other APIs which have a clear separation between `oid` and `ref` if I'm not mistaken. But that would allow us to refactor what `ref` accepts later on without deprecating the `oid` API.

mojavelinux:
All (or nearly all) of the high level functions (like checkout) in isomorphic-git accept ref rather than oid. That ref is then resolved using `GitRefManager.resolve`. I think we should match that behavior for this function since function belongs in the family of high level functions. I don't think this is the place where we should be redefining what the `ref` param means, though I do think it is a conversation we should have.

jcubic:
This is what I've had in mind update `GitRefManager.resolve` to handle all possible ref values and every function will get the new reference support.

mojavelinux:
> That ref is then resolved using GitRefManager.resolve

Usually, but not always. We could look at this again in a separate issue and determine whether there are any places it should, but does not. I see a few commands that use shortcuts to resolve the ref that could lead to inconsistent results.

SamVerschueren:
I was able to remove the `oid` input parameter, because `ref` also covers that. Thanks for pointing that out @mojavelinux 🙏 . So the thing this PR fixes now is that it does not throw on a new empty git repository and no `ref` is explicitly provided. This allows to reset files in the index in a new empty repository, which is currently not possible.

mojavelinux:
Super!

SamVerschueren:
Awesome, you guys rock 💪 🙏 !

PR #1493, opened 2022-02-20
feat(http): Added agent parameter to request function of http plugin for Node.js

### LINKED ISSUE #1491: Pass through agent to simple-get in the http plugin for Node.js
isomorphic-git does not provide support for HTTP proxies by default. However, all HTTP calls in isomorphic-git are routed through the http plugin. If the http plugin were to pass through the agent parameter to the underlying http client library (simple-get), then it would be easy to add HTTP proxy support using a custom agent. (There are plenty of such agents available in npm, including hpagent). However, since the http plugin does not pass through this parameter, it's necessary to completely replace the http plugin with a custom implementation.

Here's how it could work:

```js
'use strict'

const { request: delegate } = require('isomorphic-git/http/node')
const { HttpProxyAgent, HttpsProxyAgent } = require('hpagent')
  
async function request ({ url, method, headers, body }) {
  const proxy = url.startsWith('https:')
    ? { Agent: HttpsProxyAgent, url: process.env.https_proxy } 
    : { Agent: HttpProxyAgent, url: process.env.http_proxy }
  const agent = proxy.url ? new proxy.Agent({ proxy: proxy.url }) : undefined
  return delegate({ url, method, headers, agent, body })
}

module.exports = { request }
```

Here's the change that is needed to support this:

```diff
diff --git src/http/node/index.js src/http/node/index.js
index c4d2ff13..4d7980ac 100644
--- src/http/node/index.js
+++ src/http/node/index.js
@@ -16,6 +16,7 @@ export async function request({
   url,
   method = 'GET',
   headers = {},
+  agent,
   body,
 }) {
   // If we can, we should send it as a single buffer so it sets a Content-Length header.
@@ -30,6 +31,7 @@ export async function request({
         url,
         method,
         headers,
+        agent,
         body,
       },
       (err, res) => {
```

mojavelinux:
resolves #1491

- pass agent parameter to underlying http client
- update reference docs to include new parameter
- update faq to show how to wrap built-in http plugin to inject an agent

## I'm adding a parameter to the http plugin

- [x] add parameter to the function
- [x] add a test case in `__tests__/test-clone.js`

### DIFF
--- a/__tests__/test-clone.js
+++ b/__tests__/test-clone.js
@@ -250,4 +250,40 @@ describe('clone', () => {
     expect(merge).toBe('refs/heads/main')
     expect(remote).toBe('foo')
   })
+
+  if (typeof process === 'object' && (process.versions || {}).node) {
+    it('should allow agent to be used with built-in http plugin for Node.js', async () => {
+      const { fs, dir, gitdir } = await makeFixture('isomorphic-git')
+      const connectionLog = []
+      const { Agent } = require('https')
+      const httpWithAgent = {
+        async request({ url, method, headers, body }) {
+          const agent = new Agent()
+          // @ts-ignore
+          agent.createConnection = new Proxy(agent.createConnection, {
+            apply(target, self, args) {
+              const { hostname, port, method: method_ } = args[0]
+              connectionLog.push(`${method_} ${hostname}:${port}`)
+              return target.apply(self, args)
+            },
+          })
+          return http.request({ url, method, headers, agent, body })
+        },
+      }
+      await clone({
+        fs,
+        http: httpWithAgent,
+        dir,
+        gitdir,
+        depth: 1,
+        ref: 'test-branch',
+        singleBranch: true,
+        noCheckout: true,
+        url: 'https://github.com/isomorphic-git/isomorphic-git.git',
+      })
+      expect(connectionLog.length).not.toBe(0)
+      expect(connectionLog[0]).toEqual('GET github.com:443')
+      expect(connectionLog[1]).toEqual('POST github.com:443')
+    })
+  }
 })
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -186,7 +186,8 @@ But until then, there's no advantage to using the new protocol.
 
 _Answer by Dan Allen (@mojavelinux):_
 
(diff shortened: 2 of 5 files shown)


jcubic:
Some tests failed, maybe type checking, there is a single big command that failed (few commands with `&&`).

mojavelinux:
Argh. No, the problem is we are testing on an ancient version of Node.js, which doesn't have the same API for agent. I'll have to find another way to do the assertion.

mojavelinux:
Looks like we're in the clear now.

jcubic:
Does this test run? Or this if make it that it doesn't even run in CI? Maybe it's worth taking time to update tests and NodeJS. I'm not sure (when tests run on Azure) if we can modify the o what system they run.

mojavelinux:
Yes it runs. That's what I meant when I said "we're in the clear now". The test runs and it succeeds. I wouldn't push it if it didn't run.

mojavelinux:
https://dev.azure.com/isomorphic-git/isomorphic-git/_build/results?buildId=2561&view=logs&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=5b4cc83a-7bb0-5664-5bb1-588f7e4dc05b&l=229

> Yes it runs.

I rewrote the test to use a Node.js proxy to intercept the call to Agent.createConnection to ensure it is used. This approach works on Node.js 10. The first version of the test used an event which is only available on Node.js 16. That's why I had to rewrite the test.

mojavelinux:
Yes, we need to update Node.js in CI, but that's a separate issue. And when we do that, I really want to get ride of Azure and run the tests on GitHub Actions instead. It would take me an afternoon, but it's entirely doable. See #1496.

mojavelinux:
The new FAQ isn't appearing on the website. What is the process for pushing updates to the website? Does the command need to be invoked explicitly?

jcubic:
I'm not sure. I wanted the take a look when I found that WikiMedia image (SVG) for Android changed the name and it break the site. The repo has an updated image but the website still has the wrong one. I have no idea where that file is located. I was not searching but I think that we will need to figure out how the website is built. I will do that when I will find some free time. Unless you want to dig deep and fire that out.

The important thing is that we need to document how the build work. It may also help when we move to GitHub actions.

mojavelinux:
Thanks for that information. I will try to find some time to investigate and see what it is that we have to do. I'll comment on #1496 with anything I find out.

PR #1498, opened 2022-02-24
fix(statusMatrix): fix status matrix if entry changes type

SamVerschueren:
## I'm fixing a bug or typo

- [x] squash merge the PR with commit message "fix(statusMatrix): fix status matrix if entry changes type"

While testing things, I noticed `statusMatrix` had a bug. Basically it comes down to entries changing type, meaning something that is of type `tree` becomes a `blob` or vice versa.

For instance, you have a folder with a file `a/a.txt`. You remove the entire folder and create a file `a`. In this case, `a` was not part of the matrix because it detected `a` as being a `tree` in head.

I added tests for both scenario's. They both failed before the changes I made.


SamVerschueren:
Don't really get why it fails on Mobile Safari 🤯 .

mojavelinux:
It looks to me like at least one of the tests is failing on all browsers, not just mobile Safari.

SamVerschueren:
Mmm, interesting. I'll have a look in a bit. Might be a wrong method I call in LightningFS then.

SamVerschueren:
I think I fixed it. It was because the two tests worked on the same fixture and modified files. So the status matrix was different depending on in which order they executed. So I duplicated them and now they run in full isolation.

jcubic:
The code looks good, thanks for the fix. @mojavelinux what do you think about the fix?

mojavelinux:
I think it will meet the expectations that users have when using statusMatrix, so I'm prepared to move forward with the change.

PR #1503, opened 2022-03-03
feat: add flag to turn off fast forward merges

SamVerschueren:
This PR adds an option to `merge` and `pull` to skip fast forwards by setting `fastForward: false`. This is in line with the `--no-ff` flag in [merge](https://git-scm.com/docs/git-merge#_options) and [pull](https://git-scm.com/docs/git-pull#_options_related_to_merging).

I also noticed that there where no tests for `pull` at all, so added some together with tests for `fastForward: false`.  The changes itself our very minimal. Most of the files are fixtures.

- [x] squash merge the PR with commit message "feat: add flag to turn off fast forward merges"


mojavelinux:
Seems reasonable to me. And I appreciate the additional tests very much.

Unrelated to this particular issue, I would like to see if we could find a way to avoid having to commit all these fixture repositories. I think doing it this way makes the tests difficult to maintain. I don't see why we could have a test helper that created the repository under test dynamically. It's also easier to see what the fixture contains. If you join https://gitter.im/isomorphic-git/Lobby we could discuss it there. It shouldn't block this issue.

jcubic:
How about we mark all fixtures as binary in `.gitatributes`?

SamVerschueren:
Joined Gitter!

Also, tests seem to fail but I think they are unrelated 🤔 . I'm not able to restart them unless I commit something?

jcubic:
The tests are failing, I have no idea why. Something with Azure DevOps.

Issue #1496, opened 2022-02-21
Switch CI pipeline to GitHub Actions


mojavelinux:
It would be a lot easier to maintain the CI pipeline if it was moved to GitHub Actions. It reduces the complexity of the CI set up and is better understood by the current maintainers. Using Azure requires managing and linking an extra account rather than using the one associated with the project. Using GitHub Actions also makes it easier to perform actions on the repository that require authentication, such as publishing to GitHub Pages.

When we make this change, we should switch the Node.js version from 10 to 16 so we can start upgrading and modernizing the build process...and so the library is actually tested on a supported version of Node.js.

jcubic:
GitHub actions support matrix testing with different versions of Node. I think that we should use Node 12 or maybe Node 14 as lowers possible version that is supported. So those also need to run in CI. The end of life for 12 is in 2022-04-30. It's will be soon so maybe just Node 14 as LTS can be supported.

mojavelinux:
Agreed. We should test the range of active releases.

SamVerschueren:
I think testing all active LTS releases makes sense. So `14` and `16`?

mojavelinux:
> I don't know if we can migrate to GitHub Actions without access to NPM.

We must have that access to proceed. There is no future of this project without that.

mojavelinux:
We'll also need to figure out how the website is built and integrate it into the release pipeline.

seanpoulter:
The website isn't in the Azure Pipelines. Where's that build?
https://dev.azure.com/isomorphic-git/isomorphic-git/_build?view=folders

Anyone have access to the source code to compare before/after or are we rebuilding from the logs?

jcubic:
I have no idea it's in the azure file:

https://github.com/isomorphic-git/isomorphic-git/blob/main/azure-pipelines.yml#L79

Maybe it only runs and shows when the file in website directory changes.

seanpoulter:
Ah, I was expecting a separate pipeline. That looks like what's running. Thanks @jcubic.
![image](https://github.com/isomorphic-git/isomorphic-git/assets/2585460/82eb9646-a015-4490-baae-cda2bdbf239c)

Any other requests? We'll set it up with a Matrix so we can use the current Node v12 and other versions as we decide what to support.

jcubic:
[...]

PR #1436, opened 2021-11-03
Patch use .cjs instead of .js for test

sosuisen:
## I'm fixing a bug or typo

- [x] if this is your first time contributing, run `npm run add-contributor` and follow the prompts to add yourself to the README
- [x] squash merge the PR with commit message "fix: use .cjs instead of .js for test"

---
This is a pull request for 
#1434

PR #1506, opened 2022-03-11
Mshanemc/multi-add-from-1228

mshanemc:
This is a re-work of https://github.com/isomorphic-git/isomorphic-git/pull/1228  (@crutchcorn) to help move things along

## I'm adding a parameter *type* to an existing command add:

FYI, setup for local testing is `package.json`

``` json
  "type": "commonjs",
```


mshanemc:
basic summary of where this landed
1. promise.allSettled (can't get polyfill/shim to work) so I moved everything to node12
2. ran into the cjs/esm problems that you know well, so I made changes based on https://github.com/isomorphic-git/isomorphic-git/pull/1436 to get it to build without changing `module` to `commonjs`
3. placated the linters (I think) in one large commit to help tell the changes from the "changes"
4. had to bump some dependencies for tests to find all the modules.  Not sure what that's about.
5. had to set config file on the webpack script to get it to produce the umd in the normal place.

Once all that was done (mostly just banging away on it) I can get this to pass tests locally and in the CI.  🎉

mojavelinux:
> promise.allSettled (can't get polyfill/shim to work) so I moved everything to node12

I would prefer for the move to Node.js 12 to happen in a separate PR. If the maintainers agree with that position, would you be willing to put that together?

mojavelinux:
We could also really use help with #1496. I think that would make the contribution process a lot more straightforward and help eradicate some of these build / CI issues. Would you be willing to help with that?

jcubic:
I'm all in removing support for Node 10, there may be a problem to disable it on Azure. If I will not be able to remove it, maybe we will need to wait for GitHub actions with this.

Another idea is to just merge this as is, and later remove Node 10 when we will migrate to GitHub actions, first we will need access to isomorphic-git NPM repo, we need to be added as contributors. I'm wondering if we can just modify pakcage.json add ourselves as collaborators and on the next deployment will allow us to get NPM credentials or tokens whatever is needed to publish to NPM (I've never done this in CI).

Just thinking at loud will check what is needed for migration to GitHub actions.

jcubic:
@mshanemc it would be also a good idea to remove all prettier fixes from PR or if they actually should be there, then we will need to check why there is no bot for fixing the code anymore.

Actually, it was really nice, because we did not need to worry about style.

mshanemc:
> > promise.allSettled (can't get polyfill/shim to work) so I moved everything to node12
> 
> I would prefer for the move to Node.js 12 to happen in a separate PR. If the maintainers agree with that position, would you be willing to put that together?

I would.  That PR would have to go in ahead of what's left of adding multiple files, which is what we really need to be able to use iso-git

mshanemc:
> We could also really use help with #1496. I think that would make the contribution process a lot more straightforward and help eradicate some of these build / CI issues. Would you be willing to help with that?

It's good to do, but doesn't solve anything I encountered--the nps scripts, module vs. cjs differences between node versions, and eslint/prettier issues are more of a problem than the azure pipelines.

mshanemc:
> I'm all in removing support for Node 10, there may be a problem to disable it on Azure. If I will not be able to remove it, maybe we will need to wait for GitHub actions with this.
This PR changed Azure to run on 12.  That worked fine.

> 
> Another idea is to just merge this as is, and later remove Node 10 when we will migrate to GitHub actions, first we will need access to isomorphic-git NPM repo, we need to be added as contributors. I'm wondering if we can just modify pakcage.json add ourselves as collaborators and on the next deployment will allow us to get NPM credentials or tokens whatever is needed to publish to NPM (I've never done this in CI).
> 
> Just thinking at loud will check what is needed for migration to GitHub actions.

mojavelinux:
> It's good to do, but doesn't solve anything I encountered...

Agreed. I wasn't suggesting that effort as a prerequisite. I was roping you in since you've already gotten into the build process, I think you're well on your way to being able to help us solve it. The Azure pipelines create an incredible amount of confusion since so few people are familiar with them.

mojavelinux:
> I would. That PR would have to go in ahead of what's left of adding multiple files, which is what we really need to be able to use iso-git

I think dropping Node.js 10 needs to go in first. The alternative is to get this build working with Node.js 10. We can't do both in one commit as it makes the upgrade difficult to isolate when studying the history. It's also just not a good practice to couple discrete efforts.

mshanemc:
> @mshanemc it would be also a good idea to remove all prettier fixes from PR or if they actually should be there, then we will need to check why there is no bot for fixing the code anymore.
> 
> Actually, it was really nice, because we did not need to worry about style.

I tried to align the .prettierrc with what the project was doing already (ex: singlequotes instead of prettier's default double, no-semi) based on all the errors it was throwing.

What remained (mostly `function[space](params)`) prettied _just does_ in v2 so it's gonna happen whenever y'all upgrade anyway
https://prettier.io/blog/2020/03/21/2.0.0.html#always-add-a-space-after-the-function-keyword-3903httpsgithubcomprettierprettierpull3903-by-j-f1httpsgithubcomj-f1-josephfrazierhttpsgithubcomjosephfrazier-sosukesuzukihttpsgithubcomsosukesuzuki-thorn0httpsgithubcomthorn0-7516httpsgithubcomprettierprettierpull7516-by-bakkothttpsgithubcombakkot

mshanemc:
> > It's good to do, but doesn't solve anything I encountered...
> 
> Agreed. I wasn't suggesting that effort as a prerequisite. I was roping you in since you've already gotten into the build process, I think you're well on your way to being able to help us solve it. The Azure pipelines create an incredible amount of confusion since so few people are familiar with them.

I'd agree to a little OSS volunteer work if my PR(s) go in 😄

mshanemc:
> > I would. That PR would have to go in ahead of what's left of adding multiple files, which is what we really need to be able to use iso-git
> 
> I think dropping Node.js 10 needs to go in first. The alternative is to get this build working with Node.js 10. We can't do both in one commit as it makes the upgrade difficult to isolate when studying the history. It's also just not a good practice to couple discrete efforts.

OK, will make that a separate PR without the multi-add and will see if it's possible to use the older version of prettier and still get past the linter.

mojavelinux:
I'm not sure if this helps, especially since we're talking about dropping Node.js 10 anyway, but just in case...

Here's the polyfill I've used in the past for Promise.allSettled.

```js
Promise.allSettled = (iterable) =>
  Promise.all(
    iterable.map((it) =>
      it.then(
        (value) => ({ status: 'fulfilled', value }),
        (reason) => ({ status: 'rejected', reason })
      )
    )
  )
```

mshanemc:
node12 PR here: https://github.com/isomorphic-git/isomorphic-git/pull/1507
I think I solved the `prettier` issue, too (pointer in package.json its the config file!)

mshanemc:
this PR without the node/build stuff is https://github.com/isomorphic-git/isomorphic-git/pull/1508

PR #1507, opened 2022-03-12
node 12 and dependency/ESM updates

mshanemc:
prerequisite for https://github.com/isomorphic-git/isomorphic-git/pull/1506

1. requires node12+ (and runs azure pipeline using node12)
2. that requires changes to support ESM (many .js files become cjs)
3. node12 also requires newer jest

### for discussion
this is producing a `dist/main.js` file, which I've added to gitignore.  Not sure why it started doing that but previously hadn't 

adding prettier to the package.json seems to help it find the imported config that it was missing--that avoided having to make any prettier changes!  [it's still on v1.x]


mojavelinux:
Could you update the title of the PR so that it identifies the change being made here? "Mshanemc/node 12" is too cryptic.

jcubic:
@mshanemc The build failed when deploying the website, `require` not found. Can you take a look and create another PR? https://dev.azure.com/isomorphic-git/isomorphic-git/_build/results?buildId=2612&view=logs&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=6c939d89-0d1a-51f2-8b30-091a7a82e98c

viko16:
`engines` requirement was changed from node10 to node12, which is a BREAKING CHANGE, isn't it?

mojavelinux:
Before I get to my comment, I want to emphasize that I assist with maintaining the plugin, but I do not speak for the project.

I'm of two minds on this. Yes, I do agree that a strong case could be made that dropping a platform version is a major/breaking change. On the other hand, Node.js 10 is no longer supported and not even recognized on the Node.js [releases page](https://nodejs.org/en/about/releases/) anymore. Furthermore, the build was partially broken on Node.js 10 (for various reasons) so the library couldn't even been properly validated.

The version change here was a minor version. isomorphic-git has never had a major release since the maintainer of the library changed. To do so would be a huge deal. These minor versions are going out more or less as next versions since that's the only release process the project currently has. I would recommend pinning your version range accordingly.

jcubic:
I agree that version 2.0 should be reserved for breaking changes in the API not dropping obsolete platforms.

PR #1513, opened 2022-03-14
ci: website publishes fix

mshanemc:
consider this a repeat of #1507

There were some failures doing the website build/publish in Azure.

I tested those locally, too, and think this might work now...but it also potentially only unblocks the next build error.


mshanemc:
@jcubic maybe this one would work?

jcubic:
Yes, thanks. I think that we don't have any other option than to keep deploying and see if it work.

PR #1514, opened 2022-03-14
feat: add supports array of files

mshanemc:
This PR requires https://github.com/isomorphic-git/isomorphic-git/pull/1507

## I'm modifying a parameter to an existing command X:

### DIFF
--- a/__tests__/test-add.js
+++ b/__tests__/test-add.js
@@ -58,9 +58,11 @@ describe('add', () => {
     expect(err.caller).toEqual('git.add')
     expect(err.name).toEqual('NotFoundError')
   })
-  it('multiple files with 2 failures (MultipleGitError)', async () => {
+  it('multiple files with 2 failures (MultipleGitError) and an ignored file', async () => {
     // Setup
     const { fs, dir } = await makeFixture('test-add')
+    await writeGitIgnore(fs, dir)
+
     // Test
     await init({ fs, dir })
     let err = null
@@ -68,7 +70,7 @@ describe('add', () => {
       await add({
         fs,
         dir,
-        filepath: ['a.txt', 'non-existent', 'also-non-existent'],
+        filepath: ['a.txt', 'i.txt', 'non-existent', 'also-non-existent'],
       })
     } catch (e) {
       err = e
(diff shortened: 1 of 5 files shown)


src/api/add.js
```diff
@@ -53,27 +53,54 @@ export async function add({
 
 async function addToIndex({ dir, gitdir, fs, filepath, index }) {
   // TODO: Should ignore UNLESS it's already in the index.
-  const ignored = await GitIgnoreManager.isIgnored({
-    fs,
-    dir,
-    gitdir,
-    filepath,
+  filepath = Array.isArray(filepath) ? filepath : [filepath]
+  const promises = filepath.map(async filePathIterator => {
```
jcubic, line 57: filePathIterator it's a string, why do you call it `iterator`? I would rename it to `filepath` it will shadow the main parameter that is not needed in this context.

src/api/add.js
```diff
@@ -53,27 +53,54 @@ export async function add({
...
   }
+  if (rejectedPromises.length === 1) {
+    throw rejectedPromises[0]
+  }
+
+  const fulfilledPromises = settledPromises
+    .filter(settle => settle.status === 'fulfilled')
+    .map(settle => settle.value)
+
+  return fulfilledPromises
```
jcubic, line 105: You need to filter out undefined values. From the map:
```javascript
if (ignored) return
```

__tests__/test-add.js
```diff
@@ -36,6 +36,50 @@ describe('add', () => {
...
+    } catch (e) {
+      err = e
+    }
+    expect(err.caller).toEqual('git.add')
+    expect(err.name).toEqual('MultipleGitError')
+    expect(err.errors.length).toEqual(2)
+    err.errors.forEach(e => {
+      expect(e.name).toEqual('NotFoundError')
+    })
+  })
```
jcubic, line 82: Can you test adding a file that is ignored?

jcubic (changes requested):
Looks promising, but found one issue and I also would to see the test added for that thing.

src/api/add.js
mshanemc: I changed it to `currentFilepath` for clarity.  Not a fan of shadow.

jcubic (approved):

mojavelinux:
This change caused the fix for #1381 to be rolled back. Somehow, this change also caused the test to be removed. And it was not a simple oversight. The import of the posixifyPathBuffer function was removed, which feels very deliberate to me. There should have been an inquiry if you didn't understand why that function was being used.

We've got to be more careful about causing previous fixes to be rolled back when introducing new features.

mojavelinux:
> this change also caused the test to be removed

Correction. The test is still there. But I suppose the tests are never being run on Windows, which is when we would have seen it fail.

PR #1517, opened 2022-03-16
feat: add updateIndex API

### LINKED ISSUE #1494: Add API for `update-index --cacheinfo`
I'm looking for an API that could cover `git update-index --cacheinfo` functionality. Currently `isomorphic-git` is able to write objects to the object database with `writeObject()`, but after that it's not possible to do something with that data.

The `git update-index --cacheinfo` method allows to update the index with data that's already in the object database, but does not exist in the working directory. This is possible with `git hash-object -w --stdin` where the data that's coming from stdin gets hashed and written to the object database. The hash is returned, and can then be used to update the index with `git update-index --add --cacheinfo`.

I'm not sure how fine-grained an API should be. Should we create one API that covers a lot of the `updateIndex` functionality, or focus on `git update-index --cacheinfo` specifically (which should solve my use case).

Having a `updateIndex()` method that covers everything from `git update-index` might be a bit too much. So something specific for this use case might be better?

```ts
addCacheObject({
	// usual isomorphic-git options like fs, gitdir, dir, cache
	mode: 100644,
	object: 'b6fc4c620b67d95f953a5c1c1230aaab5db5a1b0',
	filepath: 'readme.md'
});
```

SamVerschueren:
closes #1494

I've added a new API, which is the `updateIndex` API. This is based on the [git update-index](https://git-scm.com/docs/git-update-index) command.

The current implementation allows for adding, removing, force-removing files on disk, or files which are virtual and only live in the object database. The latter use case covers `git update-index --cacheinfo <mode>,<object>,<file>`.

I'm throwing errors based on how `git update-index` throws. I was back-and-forth on adding new error types, but wasn't sure so I worked with what was already there.

I've added quite some tests to it, and tested this new API locally already with the tool that I'm building and seems to work like it should.

## I'm adding a new command:

- [x] add as a new file in `src/api` (and `src/commands` if necessary)
- [x] update `__tests__/test-exports.js`
- [x] document the command with a JSDoc comment
- [x] add page to the Docs Sidebar `website/sidebars.json`
- [x] add page to the v1 Docs Sidebar `website/versioned_sidebars/version-1.x-sidebars.json`
- [x] squash merge the PR with commit message "feat: add updateIndex API"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -130,6 +130,7 @@ docs/setConfig.md
 docs/status.md
 docs/statusMatrix.md
 docs/tag.md
+docs/updateIndex.md
 docs/version.md
 docs/walk.md
 docs/writeBlob.md
--- a/README.md
+++ b/README.md
@@ -195,6 +195,7 @@ unless there is a major version bump.
 - [status](https://isomorphic-git.github.io/docs/status.html)
 - [statusMatrix](https://isomorphic-git.github.io/docs/statusMatrix.html)
 - [tag](https://isomorphic-git.github.io/docs/tag.html)
+- [updateIndex](https://isomorphic-git.github.io/docs/updateIndex.html)
 - [version](https://isomorphic-git.github.io/docs/version.html)
 - [walk](https://isomorphic-git.github.io/docs/walk.html)
 - [writeBlob](https://isomorphic-git.github.io/docs/writeBlob.html)
--- a/__tests__/test-exports.js
+++ b/__tests__/test-exports.js
@@ -65,6 +65,7 @@ describe('exports', () => {
         "status",
         "statusMatrix",
         "tag",
+        "updateIndex",
         "version",
         "walk",
         "writeBlob",
--- a/__tests__/test-updateIndex.js
+++ b/__tests__/test-updateIndex.js
@@ -1,0 +1,268 @@
+/* eslint-env node, browser, jasmine */
+const path = require('path')
+
+const { writeBlob, updateIndex, status, add } = require('isomorphic-git')
+
+const { makeFixture } = require('./__helpers__/FixtureFS.js')
+
+describe('updateIndex', () => {
+  it('should be possible to add a file on disk to the index', async () => {
+    // Setup
+    const { fs, dir } = await makeFixture('test-empty')
+    await fs.write(path.join(dir, 'hello.md'), 'Hello, World!')
+    // Test
+    const oid = await updateIndex({
+      fs,
+      dir,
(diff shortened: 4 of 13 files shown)


jcubic (changes requested):
Please if possible add filepath as string and array of strings. This is a recent change (I think in git.add) and it would be nice to have the same in this API.
**question**: how it will work when the file is in gitigore? Does the update index is affected by gitignore? If so then please add proper code (you can check code for git add) and unit test for gitignore.

SamVerschueren:
The array filepath list is pretty clear when removing files. But I'm not sure what to do for adding and could use some feedback.

Just adding regular files is pretty clear as well. (ignoring common properties like `fs` for simplicity)

```ts
await git.updateIndex({
    add: true,
    filepath: ['readme.md', 'package.json']
});
```

But when passing in an `oid`, it doesn't really make sense to provide multiple file paths.

```ts
await git.updateIndex({
    add: true,
    filepath: ['readme.md', 'package.json'],
    oid: '...'
});
```

I think we might have 3 options here
1. We throw an error. (Not a big fan of this one)
2. We just allow it, and then the same object from the object database will be added 3 times.
3. We remove `oid` and `mode` from the root object and allow `string|object|(string | object)[]` as `filepath`.

The 3rd option would then look like this

```ts
await git.updateIndex({
    add: true,
    filepath: [
		{
			filepath: 'readme.md',
			oid: '...'
		}
        'package.json'
	]
});
```

If we go with this, I see 2 (small) issues.
1. `filepath` is not really a good property key, I guess it should be renamed to `files` then or something
2. Defining this in JSDoc seems to be impossible. It seems complains about nested properties in this scenario.

```
 * @param {string|string[]|object} args.filepath - File to act upon.
 * @param {string} args.filepath.oid -
```

image

jcubic:
If it doesn't make sense to use an array with oid, then just throw an error. But it should be documented that array works only for adding.

jcubic:
I'm thinking, the reason why filepath was an array was that adding each file was really slow. Maybe if updateIdndex doesn't have such problems, then maybe we should leave it as is. Using single string. Especially if it's tricky. What do you think?

SamVerschueren:
I'm wondering if `add` was slow because it constantly had to parse the ignore rules for every add you do 🤔 . That is not the case for `updateIndex`. I'm definitely ok with leaving it as is if you are all ok with this.

jcubic (approved):

jcubic:
If so I'm merging the PR, thanks.

Issue #1515, opened 2022-03-14
ignoring gitignore


mshanemc:
I want to be able to "ignore the .gitignore"
git add --force 

I'd like to implement that, along with the equivalent of `status --ignored` for `statusMatrix`.

jcubic:
Go ahead.

PR #1519, opened 2022-03-19
feat: add --force and statusMatrix --ignored

mshanemc:
## I'm adding a parameter to an existing command add/statusMatrix:
see https://github.com/isomorphic-git/isomorphic-git/issues/1515

create a boolean `force` param for add (equivalent of `git add --force`...add it to the index even if it would match the .gitignore)

create a boolean `ignored` param for statusMatrix (it already returns anything in the index, even if ignored, but would NOT show new files if they are ignored).  This allows for that use case

---

> note 1: the `git status --ignored` flag was at some point modfied to take arguments.  

This hard to implement as an object with a default, so I left it as a boolean and it's doing "traditional"
see https://git-scm.com/docs/git-status

My recommendation would be that if you ever wanted those, you'd have another prop on the object called ignored-mode that defaults to "traditional" and implements all that other logic, so that `ignored` would represent the "presence" of the `ignored` flag, and the ignored-mode would represent its value

---
> note 2: the status command returns `ignored` for ignored files, which seems perfectly fine

### DIFF
--- a/src/api/statusMatrix.js
+++ b/src/api/statusMatrix.js
@@ -160,7 +160,7 @@ export async function statusMatrix({
   filepaths = ['.'],
   filter,
   cache = {},
-  ignored = false,
+  ignored: shouldIgnore = false,
 }) {
   try {
     assertParameter('fs', _fs)
@@ -177,16 +177,15 @@ export async function statusMatrix({
       map: async function(filepath, [head, workdir, stage]) {
         // Ignore ignored files, but only if they are not already tracked.
         if (!head && !stage && workdir) {
-          if (
-            ignored
-              ? false
-              : await GitIgnoreManager.isIgnored({
-                  fs,
-                  dir,
-                  filepath,
-                })
-          ) {
-            return null
+          if (!shouldIgnore) {
+            const isIgnored = await GitIgnoreManager.isIgnored({
+              fs,
+              dir,
+              filepath,
+            })
+            if (isIgnored) {
+              return null
+            }
           }
         }
         // match against base paths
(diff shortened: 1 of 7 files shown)


mshanemc:
@jcubic got this to 🟢.

src/api/statusMatrix.js
```diff
@@ -176,11 +178,13 @@ export async function statusMatrix({
...
-              filepath,
-            })
+            ignored
+              ? false
+              : await GitIgnoreManager.isIgnored({
+                  fs,
+                  dir,
+                  filepath,
+                })
           ) {
```
jcubic, line 188: I would use:

```
ignored: shouldIgnore = false,
```   

```javascript
if (!head && !stage && workdir) {
   if (!shouldIgnore) {
      const isIgnored = await GitIgnoreManager.isIgnored({
        fs,
        dir,
        filepath,
      });
      if (isIgnored) {
         return null
      }
   }
}
```

jcubic: Also while you're editing this file, can you check (and maybe write a test) for a status matrix with an array of strings? This looks like it only works the first single `filepath`.

jcubic (changes requested):
Can you add those changes, in the comments?

As for ignored in statusMatrix maybe something like:

```javascript
const statusMode = shouldIgnore === shouldIgnore ? "traditional" : shouldIgnore;
```

But this can be done later, if something will want to do this.

mshanemc:
> ```js
> const ignoreMode = shouldIgnore === true ? "traditional" : shouldIgnore;
> ```
> 
> But this can be done later if something will want to do this.
I'm not sure about that design but going to defer anyone--people may or may not need it.

If I understood about the 'array of strings' comment, I couldn't reproduce your concern (upside: there's more tests about it)

jcubic (approved):
Looks Good.

PR #1520, opened 2022-03-21
fix: release step

SamVerschueren:
I noticed that the builds are failing during the release step, which causes the website not to be updated with the latest documentation. After looking at the logs, I saw that it tries to run `tweet.js` as an ES module. Changed it to be `CJS` hoping that it will fix the problem. I tried running it like this locally and at lease it runs. It fails on the tokens though, but that's expected on my local machine.

## I'm fixing a bug or typo

- [x] squash merge the PR with commit message "fix: release step"


SamVerschueren:
I'm not sure if it fixes the entire build. But it definitely fixes the tweet issue.

jcubic:
Thanks

jcubic:
I think it's working updateIndex is on the website.

PR #1530, opened 2022-04-05
fix: Assert parameter is confusing... fix

rayfoss:
## I'm fixing a bug or typo

```
MissingParameterError: The function requires a "filepaths" parameter but none was provided.
```

There is no such parameter as filepaths... there is a filepath, hence the PR. 

Also, we are not following any case convention :/ , corsProxy and singleBranch are came cased, but filepath is lower cased. I propose deprecating and printing a deprecation warning.

To make matters worse filePath is not a normal file path... a path relative to the git repo. perhaps it should be called realtivePath instead.

### DIFF
--- a/src/api/add.js
+++ b/src/api/add.js
@@ -41,7 +41,7 @@ export async function add({
     assertParameter('fs', _fs)
     assertParameter('dir', dir)
     assertParameter('gitdir', gitdir)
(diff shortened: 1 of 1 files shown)


jcubic:
I think that the fix is not added because "fix" was first in the title and semantic release didn't pick it up.

PR #1532, opened 2022-04-13
chore(deps-dev): bump semantic-release and @semantic-release/exec

seanpoulter:
## I'm updating a dependency

This PR updates `semantic-release` to close #1266 and `@semantic-release/exec` to close #1086. This also fixes the problems reported using npm 7 or later about peer dependencies:

Screen Shot 2022-04-12 at 21 19 05

This is the furthest we can upgrade with Node.js v12:

- `semantic-release@>=18` requires Node.js v14.
  See https://github.com/semantic-release/semantic-release/releases/tag/v18.0.0.

- `@semantic-release/exec@>=6` requires Node.js v14.  
  See https://github.com/semantic-release/exec/releases/tag/v6.0.0.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -558,6 +558,15 @@
         "doc",
         "test"
       ]
+    },
+    {
+      "login": "seanpoulter",
+      "name": "Sean Poulter",
+      "avatar_url": "https://avatars.githubusercontent.com/u/2585460?v=4",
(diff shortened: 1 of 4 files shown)


jcubic:
Thanks for the fix. Not sure if the bot will release a new version right now, we probably will need to wait for the next fix or feature to see if the changes are correct. Sometimes things work locally but not on CI/CD.

PR #1561, opened 2022-04-25
draft: Update karma-sauce-launcher

seanpoulter:
This is a work in progress. Perhaps it will help our Android and iOS test stability.


seanpoulter:
Feel free to unsubscribe from this thread for a while @jcubic. I'll send you a message here and on Gitter once it is good to go. There will be some trial and error here starting from one browser and building up.

Oh - and I've learned that as of January our current style of configuring browsers is deprecated. Here's a reference article - https://support.saucelabs.com/hc/en-us/articles/4412359870231-Migrating-Appium-Real-Device-Tests-to-W3C.

Issue #1256, opened 2020-10-30
git config fails when config path (tracking branch) is mixed case


cvarjao:
I am trying to retrieve the remote tacking branch name, but it is failing because of mixed casing in the branch name
I was able to reproduce by adding the following scenario to [test-GitConfig.js](https://github.com/isomorphic-git/isomorphic-git/blob/v1.7.8/__tests__/test-GitConfig.js):
```
    it('section with mixed case name', async () => {
      const config = GitConfig.from(`[branch "feature/SOMETHING-123"]
      merge = valaaa`)
      const a = await config.get('branch.feature/SOMETHING-123.merge')
      expect(a).toEqual('valaaa')
    })
```
I've tracked it down to a a possible bug in the `toLowerCase()` added to only one side of the comparison:
https://github.com/isomorphic-git/isomorphic-git/blob/v1.7.8/src/models/GitConfig.js#L155

Please be sure to mention:
- whether you are using Node or the Browser
- how you are using it (if you are using a bundler what bundler; if you are using a `<script>` tag what CDN URL)

jcubic:
Hi, sorry for the late reply, I'm a new maintainer with admin rights but I don't think I will find time to contribute with code.

if you know the issue and found the bu, can you create PR with a fix?

ronran1:
Hi, I found this old thread and fixed the test case by adding toLowerCase() to both sides of the string. I'm opening a pull request to fix this now.

jcubic:
@ronran1 I don't think that simple toLowerCase will fix the issue, here. Your code will break how git works. Since git branches are case sensitive. The fix is not adding another toLowerCase but removing the existing one.

SamVerschueren:
This should be fixed by #1573

SamVerschueren:
This scenario is covered by [this test](https://github.com/isomorphic-git/isomorphic-git/blob/main/__tests__/test-GitConfig.js#L40-L47) now. I also ran the test case provided above, just to be sure, and it works fine. 

@jcubic I think we can close this issue now.

PR #1568, opened 2022-05-01
fix: fixed mixed case config path bug and added test case to tests for 

ronran1:
future reference. can remove if undesired. 

https://github.com/isomorphic-git/isomorphic-git/issues/1256 is the relevant issue.


jcubic:
This StackOverflow answer says that Git branches are case sensitive:

[Git branch name - case sensitive or insensitive?](https://stackoverflow.com/questions/38493543/git-branch-name-case-sensitive-or-insensitive)

So your PR is invalid. Did you test it in canonical git? If it works are you think it does?

jcubic:
I'm closing this PR since this breaks git.

PR #1573, opened 2022-05-10
fix(config): keep correct casing and support comment tokens in value

### LINKED ISSUE #1570: Writing to git config should keep casing
While I was testing [`addRemote`](https://isomorphic-git.org/docs/en/addRemote), I noticed that if I passed in a remote name like `FooBar`, it ended up as `[remote "foobar"]` in the git config.

Look at the [`GitConfig` model](https://github.com/isomorphic-git/isomorphic-git/blob/main/src/models/GitConfig.js#L207-L211), I noticed that `toLowerCase()` is used to generate the section path. I believe this is incorrect, and it should be written as it comes in.

For instance, if you run `git config FOO.bar bar` or `git config add MyRemote foo`. Looking in the `.git/config` file, the casing is intact.

Retrieving that value is case insensitive though. So running `git config foo.bar` or `git config FOO.BAR` both result in `bar`.

Now, why is this a problem? If you add an upstream by running `git remote add MyUpstream https://github.com/isomorphic-git/isomorphic-git`, casing does make a difference. In this case, I should be able to run `git fetch MyUpstream`, but not `git fetch myupstream`. For this to work correctly, the section in the `.git/config` should be `[remote "MyUpstream"]` and not `[remote "myupstream"]`. Because in the latter, the `listRemotes()` will return `myupstream` which is incorrect.

So long story short. Writing sections to the config file should not be lowercased (because casing matters in certain scenario's). Reading sections should be case insensitive.

I can send a PR to address this.

SamVerschueren:
fixes #1570 

This PR keeps the casing of the section and the name when writing the git config, but allows when retrieving a value, the section and name are still case insensitive. Just like it works with `git config`.

I also noticed that writing a value with `#` or `;` (which are comment tokens) didn't work. Basically, it wrote the value as `key = value#1` instead of `key = "value#1"`. This PR addresses that issue as well.

I added tests which test this new functionality, and all other tests still work.

## I'm fixing a bug or typo

- [x] squash merge the PR with commit message "fix(config): keep correct casing and support comment tokens in value"

### DIFF
--- a/src/models/GitConfig.js
+++ b/src/models/GitConfig.js
@@ -112,6 +112,21 @@ const getPath = (section, subsection, name) => {
     .join('.')
 }
 
+const normalizePath = path => {
+  const pathSegments = path.split('.')
+  const section = pathSegments.shift()
+  const name = pathSegments.pop()
+  const subsection = pathSegments.length ? pathSegments.join('.') : undefined
+
+  return {
+    section,
+    subsection,
+    name,
+    path: getPath(section, subsection, name),
+    sectionPath: getPath(section, subsection, null),
+  }
+}
+
 const findLastIndex = (array, callback) => {
   return array.reduce((lastIndex, item, index) => {
     return callback(item) ? index : lastIndex
@@ -151,8 +166,9 @@ export class GitConfig {
   }
 
   async get(path, getall = false) {
+    const normalizedPath = normalizePath(path).path
     const allValues = this.parsedConfig
-      .filter(config => config.path === path.toLowerCase())
+      .filter(config => config.path === normalizedPath)
       .map(({ section, name, value }) => {
         const fn = schema[section] && schema[section][name]
         return fn ? fn(value) : value
@@ -182,9 +198,16 @@ export class GitConfig {
   }
 
   async set(path, value, append = false) {
+    const {
+      section,
+      subsection,
+      name,
+      path: normalizedPath,
+      sectionPath,
+    } = normalizePath(path)
     const configIndex = findLastIndex(
       this.parsedConfig,
-      config => config.path === path.toLowerCase()
+      config => config.path === normalizedPath
     )
     if (value == null) {
       if (configIndex !== -1) {
@@ -193,7 +216,9 @@ export class GitConfig {
     } else {
       if (configIndex !== -1) {
         const config = this.parsedConfig[configIndex]
+        // Name should be overwritten in case the casing changed
         const modifiedConfig = Object.assign({}, config, {
+          name,
           value,
           modified: true,
         })
@@ -203,13 +228,6 @@ export class GitConfig {
           this.parsedConfig[configIndex] = modifiedConfig
         }
       } else {
-        const pathSegments = path.split('.')
-        const section = pathSegments.shift().toLowerCase()
-        const name = pathSegments.pop()
-        const subsection = pathSegments.length
-          ? pathSegments.join('.').toLowerCase()
-          : undefined
-        const sectionPath = subsection ? section + '.' + subsection : section
         const sectionIndex = this.parsedConfig.findIndex(
           config => config.path === sectionPath
         )
@@ -219,7 +237,7 @@ export class GitConfig {
           name,
           value,
           modified: true,
-          path: getPath(section, subsection, name),
+          path: normalizedPath,
         }
         if (SECTION_REGEX.test(section) && VARIABLE_NAME_REGEX.test(name)) {
           if (sectionIndex >= 0) {
@@ -231,7 +249,7 @@ export class GitConfig {
               section,
               subsection,
               modified: true,
-              path: getPath(section, subsection, null),
+              path: sectionPath,
             }
             this.parsedConfig.push(newSection, newConfig)
           }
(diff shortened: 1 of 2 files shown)


jcubic:
I'm not exactly sure how this works, can you explain?

The code still uses `getPath` that uses lowercase on the first and last part of the path. Which is exactly the same as before. How does this solve anything? Do tests fail if you remove your code and only left the tests?

jcubic:
Ah, sorry I just assume that the fix was for case compare, not about comments in files.

LGTM.

jcubic:
No, I'm still confusing, was the previous code correct with lowercase?

SamVerschueren:
Hi @jcubic, thanks for asking. First of all, yes the tests do fail if I remove the changes I made.

image

Let me try to explain in text what happens first, and I'll make some comments in the code as well to point some things out. 
So to be clear, this PR addresses two issues.

### 1. Comment tokens in values

This is the most easy one and I guess you already saw what this was about, but let me quickly explain it in more detail.

When running `git config branch.main.foo foo#bar`, the git config section `[branch "main"]` will get updated with `foo = "foo#bar"`. Because `#` denotes a comment in the config file, it has to be wrapped in double quotes. Same goes for `;`. Before this PR, this was not correctly serialized and it wrote `foo = foo#bar` instead.

### 2. Casing

Casing in git config feels kinda weird. But here's how it works. 

We basically have to split the use case up in retrieving from and writing to a config file.

#### Retrieving

The section and key are case insensitive when retrieving. The subsection is case sensitive when retrieving.

So let's assume the following config file

```
[foo "bar"]
    baz = foobarbaz
```

The section is `foo`, the subsection is `bar`, and the key is `baz`. Casing is not important for `foo` and `baz`, but it is for `bar`.

The config can be retrieved with `git config foo.bar.baz` or `git config FOO.bar.BAZ`, but not with `git config FOO.BAR.BAZ` because the subsection is `bar`, not `BAR` and the subsection is case sensitive.

#### Writing

This is where I noticed casing was not handled correctly. Basically, the input path was always entirely lowercased before writing the change to the config file.

So for instance, if we would do `git config FOO.BAR.BAZ foobarbaz` with isomorophic-git, it ended up in the config file like this.
```
[foo "bar"]
    baz = foobarbaz
```

In theory, this does not matter for `foo` and `baz`, because when retrieving they are case insensitive. But it is important for the subsection `bar`. If `BAR` is used, it should be retrieved with `BAR`, not `bar`.

The reason why `getPath()` is still used, is for matching. When the config file is parsed, the config items have a `path` property was is constructed with `getPath()` as well. This means that if we want to match our input with the data in the config file, we also have to use `getPath()` to construct it because then the `section` and `key` are always lowercased (basically making them case insensitive). The subsection is left alone and is used as is.

But also important to note is that the `normalizePath()` method returns the tokens as they where received without lowercasing them first (which is what happened before). We use the `path` and `sectionPath` properties to find the correct section, and then we use the other properties to write or update the correct section.

src/models/GitConfig.js
```diff
@@ -203,13 +228,6 @@ export class GitConfig {
           this.parsedConfig[configIndex] = modifiedConfig
         }
       } else {
-        const pathSegments = path.split('.')
-        const section = pathSegments.shift().toLowerCase()
-        const name = pathSegments.pop()
-        const subsection = pathSegments.length
-          ? pathSegments.join('.').toLowerCase()
-          : undefined
-        const sectionPath = subsection ? section + '.' + subsection : section
```
SamVerschueren, line 212: Let's assume we want to write `FOO.BAR.BAZ = foobarbaz`

With the old code, the values of the variables look like this
```js
pathSegments: ['FOO', 'BAR', 'BAZ']
section: 'foo'
name: 'BAZ'
subsection: 'bar'

sectionPath = 'foo.bar'
path: 'foo.bar.baz'
```

With the new `normalizePath` method, the values look like this

```js
pathSegments: ['FOO', 'BAR', 'BAZ']
section: 'FOO'
name: 'BAZ'
subsection: 'BAR'

sectionPath = 'foo.BAR'
path = 'foo.BAR.baz'
```

The `sectionPath` and `path` are used to do "case-insensitive" matching against the data in the config file (but with correct casing for the subsection). The other variables are used to write to the config file. As you can see in the `normalizePath` method, the casing is kept intact for the data like `section`, `subsection`, and `name`, which is used to write to the file.

src/models/GitConfig.js
```diff
@@ -151,8 +166,9 @@ export class GitConfig {
   }
 
   async get(path, getall = false) {
+    const normalizedPath = normalizePath(path).path
     const allValues = this.parsedConfig
-      .filter(config => config.path === path.toLowerCase())
+      .filter(config => config.path === normalizedPath)
```
SamVerschueren, line 171: The `path` property in the `config` object is also the normalized path where the `section` and `name` are lowercased, but not the `subsection`.

So if we have this config file

```
[Foo "BAR"]
    BAZ = foobarbaz
```

The `config.path` will be `foo.BAR.baz`. Whatever `path` we provide to this method, we want to normalize that input to lowercase the section and name. So if we provide `FOO.BAR.BAZ`, the normalized path will be `foo.BAR.baz` matching the value in the config file.

jcubic:
Thanks for the explanations, this seems right.

Issue #325, opened 2018-07-10
How to deal with merge conflicts


zhaopengme:
Isomorphic-git can handle merges without conflicts, but there are conflicting merges. I don't have a way. Maybe there is a way to accept one of them.

wmhilton:
Not yet, sorry!

virus2016:
Hey @wmhilton,

Any update on this one?

ralismark:
Looks like isomorphic-git already adds the merge conflict markers (in [mergeFile](https://github.com/isomorphic-git/isomorphic-git/blob/b102e1d9a8f48ec20042be5ff0feb9961007e904/src/utils/mergeFile.js#L29-L40)), but [immediately throws if they are added](https://github.com/isomorphic-git/isomorphic-git/blob/b102e1d9a8f48ec20042be5ff0feb9961007e904/src/utils/mergeTree.js#L228-L231).

Vinzent03:
Would love to see progression on this, because this is really essential for my application. Especially if it looks like part of the implementation is already done.

eugeniol:
> Looks like isomorphic-git already adds the merge conflict markers (in [mergeFile](https://github.com/isomorphic-git/isomorphic-git/blob/b102e1d9a8f48ec20042be5ff0feb9961007e904/src/utils/mergeFile.js#L29-L40)), but [immediately throws if they are added](https://github.com/isomorphic-git/isomorphic-git/blob/b102e1d9a8f48ec20042be5ff0feb9961007e904/src/utils/mergeTree.js#L228-L231).

Did you find a workaround? I'll love to contribute to this

araknast:
I've drafted a potential solution for this, @jcubic let me know what you think.

jcubic:
Will check the PR on Monday, I'm not at home and I don't want to do review on a phone.

stoyanov-ivelin:
Hey!

I'm getting ```MergeNotSupportedError: Merges with conflicts are not supported yet.``` with the code below. (basically a copy+paste from the docs)

```javascript
    const result = await git
      .merge({
        fs: fileSystem,
        dir: "/tutorial",
        ours: "main",
        theirs: "branch3",
        author: {
          name: "Ivelin",
        },
        abortOnConflict: false,
      })
      .catch((e) => {
        if (e instanceof Errors.MergeConflictError) {
          console.log(
            "Automatic merge failed for the following files: " +
              `${e.data}. ` +
[...]

PR #1588, opened 2022-05-27
feat(merge): Added 'mergeDriver' parameter 

### LINKED ISSUE #841: support incomplete merges with merge conflicts
It would be nice if isomorphic-git's `merge` could update the working directory and `.git/index` identically to how canonical git does when there are merge conflicts. I mean, no one likes manually resolving merge conflicts. But it's better than not being able to merge at all.

araknast:
Adds an optional `onMergeConflict` parameter to `merge()`. 
The parameter is a callback which takes the string contents of two conflicting files as input and returns the merged text. If no such parameter is provided, and `dir` is set when calling merge, markers are added to the conflicting files in the working directory and a `MergeNotSupportedError()` is thrown, aborting the merge.

Removes configuration for marker size and output type with their corresponding tests, as these are now redundant. 

Should fix #841 and #325 


jcubic:
Thanks for the PR with a fix, will check it on Monday.

__tests__/test-merge.js
```diff
@@ -564,7 +569,135 @@ describe('merge', () => {
...
+      onMergeConflict: ({
+        ourContent,
+        baseContent,
+        theirContent,
+        ourname = 'ours',
+        baseName = 'base',
+        theirName = 'theirs',
+      }) => {
+        const LINEBREAKS = /^.*(\r?\n|$)/gm
+        const ours = ourContent.match(LINEBREAKS)
```
jcubic, line 612: What this line is doing? It's hard to understand. Is it the same as `ourContent.split(/\r?\n/)`?

jcubic:
How about if there is more than one conflicted file? This should be handled as well. I think that the callback function (event) should be called on each file also it should return a promise. Or there should be an exception thrown and the user should call the function resolve error.

This should work so I can make this flow work like in canonical git:

merge -> conflict -> manual edit the file -> continue merge.

I would prefer the API that spits conflict and resolver into two functions for example:

```javascript
git.merge({...args, conflictMarkers: true});
// manual edit
git.resolve({...args});
```

There can be two APIs one with a callback function (event) and one when the user can code and do manual resolving of the conflict.

__tests__/test-mergeFile.js
```diff
@@ -29,33 +29,4 @@ describe('mergeFile', () => {
...
-    const { cleanMerge, mergedText } = mergeFile({
-      ourContent: await fs.read(`${dir}/a.txt`, 'utf8'),
-      baseContent: await fs.read(`${dir}/o.txt`, 'utf8'),
-      theirContent: await fs.read(`${dir}/c.txt`, 'utf8'),
-      format: 'diff3',
-      markerSize: 10,
-    })
-    expect(cleanMerge).toBe(false)
-    expect(mergedText).toEqual(await fs.read(`${dir}/aoc3-m10.txt`, 'utf8'))
-  })
```
jcubic, line 60: If you removed those two tests does it mean that this API no longer works? This would be a breaking change.

__tests__/test-merge.js
araknast: This regex is copied from the original `mergeFile` implementation, but yes it's essentially equivalent to `ourContent.split(/\r?\n/)`

__tests__/test-mergeFile.js
araknast: Options for diff3 and custom marker size were only exposed internally, and never used by anything except these tests, so this wouldn't be a breaking change.

araknast:
> How about if there is more than one conflicted file? And you need to do different action than just prefer ours or theirs. This should be handled as well.
The `prefer ours` and `prefer theirs` tests are just simple examples meant to check that `onMergeConflict` is implemented properly. The user could implement any number of strategies provided they are the same regardless of file. If the user wants to resolve conflicts differently for different files they can forgo specifying `onMergeConflict`, allow the files in the working directory to be marked, resolve the conflicts manually and then add and commit. Alternatively, we could add a `filepath` parameter to `onMergeConflict` so the conflict resolver could take different actions for different files.

> This should work so I can make this flow work like in canonical git:
> 
> merge -> conflict -> manual edit the file -> continue merge.
This is essentially the behavior when `onMergeConflict` is not specified. After the error is thrown, the user can edit the marked files in the working directory, add and commit just like in canonical git. I was incorrect in stating a `MergeNotSupportedError()` aborts the merge, it simply interrupts the merge, and writes all merged and conflicting files to the working directory similarly to canonical git. 

> There can be two APIs one with a callback function (event) and one when the user can code and do manual resolving of the conflict.
I'm not sure how this is different from what's implemented here. `merge({...args, onMergeConflict})` essentially functions as the callback api while `merge({...args})` functions at the manual resolution api. Instead of 
```
git.merge({...args, conflictMarkers: true});
// manual edit
git.resolve({...args});
```
We do
git.merge({...args})
.catch(e => {
  if (!e instanceof MergeNotSupportedError) throw e;
})
// manual edit
git.add({...args})
git.commit({...args})
```
Although I do think we need a new type of error to throw, e.g. `MergeConflictError()` to differentiate merge conflicts that can be resolved manually from those that can't. 

In the former example, assuming `git.merge()` doesn't throw an error when a conflict occurs and `conflictMarkers` is `true`, I'm not sure how we determine when a merge is interrupted and therefore needs manual resolution. Assuming `git.merge()` does throw an error, the flow is the same except we just use the existing `git.add()` and `git.commit()` instead of adding a new `git.resolve()`.

**Comment by jcubic** [comment]
Ok, thanks for your work and explanation. Will this work?

```javascript
git.merge({...args, conflictMarkers: true})
.catch(e => {
  if (e instanceof MergeConflictError) {
     // conflict marker exists I can edit and resolve conflicts by hand
     
  }
});

git.merge({...args, conflictMarkers: false})
.catch(e => {
  if (e instanceof MergeConflictError) {
     // conflict happen merge aborted
     
  }
});
```

It would be nice to see unit tests for this scenario.

Another missing feature is the abort of the merge. If we have conflicts and the merge was stopped in the middle we need to be able to abort the merge.

araknast:
This flow should now be possible with the `clean` parameter, e.g:

```
git.merge({...args})
.catch(e => {
  if (e instanceof MergeConflictError) {
     // conflict marker exists I can edit and resolve conflicts by hand
  }
});

git.merge({...args, clean: true})
.catch(e => {
  if (e instanceof MergeConflictError) {
     // conflict happen merge aborted - working and stage area clean
  }
});
```
Implementing merge abortion is a bit trickier, as it seems the `GitIndex` structure isomorphic-git currently uses is implemented using `Map()`, so since each file path is permitted only one entry in the index, we can't have the index stages required for proper `merge --abort` behavior. I'll see what I can do, hopefully it's fairly trivial to fix.

araknast:
I agree, I think `abortOnConflict` works better, and tests should be working now that I've brought my repo up to date with upstream. 

I've gone ahead and added some documentation for the new parameter, and also changed the default behavior to throw an error by default to match with what we do currently.

araknast:
@jcubic since we are leaving merge --abort as a separate issue, is there anything else needed to get this merged?

jcubic:
No, sorry I need to look through your PR again since you've added a lot of commits. I simply didn't have time to check the code again.

src/utils/mergeFile.js
```diff
@@ -2,26 +2,27 @@ import diff3Merge from 'diff3'
...
+  const theirContent = contents[2]
+
   const ours = ourContent.match(LINEBREAKS)
   const base = baseContent.match(LINEBREAKS)
   const theirs = theirContent.match(LINEBREAKS)
 
   // Here we let the diff3 library do the heavy lifting.
   const result = diff3Merge(ours, base, theirs)
 
+  const markerSize = 7
```
jcubic, line 20: I would not get rid of this option and use:

```javascript
 mergeFile({ branches, contents, markerSize = 7 }) {

})
```

Just in case, so we don't remove something that may be needed.

jcubic:
I've added one comment, why did you remove markerSize? It seems that it may be useful to not remove functionality unless there is a good reason to remove that option.

The second thing is can you add both mergeDrives codes from unit test to documentation? Prefer ours and prefer theirs.

jcubic:
Ok, so maybe there should be just one unit test and one documentation piece that show how to produce conflict like in canonical git where you create the markers yourself in the driver. So you can edit the files and add a merge commit. I'm not sure about merge commit but there should be a state where there is conflict in files with markers.

jcubic:
And if you think that `markerSize` is not needed you can remove it, unless it will work when you don't provide the MergeDriver and you will end up with conflict in files.

araknast:
Ok, I've added tests and documentation for manual resolution of merge conflicts, and re-removed the markerSize parameter. The documentation also explains the current limitations of the merge command, and how to emulate the behavior of `git merge --abort` and `git merge --continue` since we don't have a proper index model yet.

jcubic:
But what about markers in the output files? You just added all files and commit after conflict.
You've said that there will be markers that users can add in mergeDriver. How he can edit the files and resolve the conflicts if there are no markers?

araknast:
If no merge driver is specified, the default merge driver (mergeFile) is used, which automatically adds markers to the files with conflicts.

jcubic:
You've said that makerSize is not needed because you write them by hand. Now you say that to have a conflict with markers you don't use the driver. If that is the case then you need to add back markerSize so you can have the same output as before.

araknast:
If the user wanted to change the marker size they can simply copy the code from `mergeFile` and change the `markerSize` variable there, e.g.

```
const mergeDriver = ({ contents, branches }) => {
  const ourName = branches[1]
  const theirName = branches[2]

  const baseContent = contents[0]
  const ourContent = contents[1]
  const theirContent = contents[2]

  const ours = ourContent.match(LINEBREAKS)
  const base = baseContent.match(LINEBREAKS)
  const theirs = theirContent.match(LINEBREAKS)

  const result = diff3Merge(ours, base, theirs)

  // const markerSize = 7
  const markerSize = 28 // change marker size

  let mergedText = ''
  let cleanMerge = true

  for (const item of result) {
    if (item.ok) {
      mergedText += item.ok.join('')
    }
    if (item.conflict) {
      cleanMerge = false
      mergedText += `${'<'.repeat(markerSize)} ${ourName}\n`
      mergedText += item.conflict.a.join('')

      mergedText += `${'='.repeat(markerSize)}\n`
      mergedText += item.conflict.b.join('')
      mergedText += `${'>'.repeat(markerSize)} ${theirName}\n`
    }
  }
  return { cleanMerge, mergedText }
}

git.merge({...args, mergeDriver})
.catch(e => {
  if (!(e instanceof MergeConflictError)) throw e
})
// manual resolution
git.add({...args})
git.commit({...args})
```

Should I add this to the documentation?

jcubic:
Yes, it would be nice to have this in the documentation.

araknast:
Ok, I've added an example to the mergeDriver documentation.

jcubic:
Thank you

PR #1599, opened 2022-06-15
fix(merge): ensure correct value of cleanMerge

araknast:
This should fix an issue where `mergeTree` might report a clean merge even if there was not one. If the most recent merged blob was merged cleanly, `cleanMerge` should still be `false` if previous blobs failed.

### DIFF
--- a/src/utils/mergeTree.js
+++ b/src/utils/mergeTree.js
@@ -118,7 +118,7 @@ export async function mergeTree({
               theirName,
               mergeDriver,
             }).then(r => {
-              cleanMerge = r.cleanMerge
+              cleanMerge = cleanMerge && r.cleanMerge
               unmergedFiles.push(filepath)
               return r.mergeResult
             })


jcubic:
Can you explain more? Can the issue be unit tested?

araknast:
Essentially if `mergeTree` finishes processing a cleanly merged file *after* processing a file with merge conflicts, it will report no merge conflicts being found, and tell `_merge` it is safe to commit the merge. 

Writing a unit test would be difficult since the bug is a result of a race condition that's unlikely to occur. It happens to work out that files with merge conflicts take the longest to process, and since the processing runs asynchronously, functions processing conflicted files will be the last to set the `cleanMerge` variable. Changing `cleanMerge = cleanMerge && r.cleanMerge` ensures that cleanMerge will always be set to `false` if any of the files have conflicts, no matter what order they're processed in.

jcubic:
Ok, thanks for the explanation.

PR #1614, opened 2022-07-21
feat: Added prune & pruneTags parameter to pull command

rraab-dev:

- [x ] add parameter to the function in `src/api/X.js` (and `src/commands/X.js` if necessary)
- [x ] document the parameter in the JSDoc comment above the function
- [x ] if this is your first time contributing, run `npm run add-contributor` and follow the prompts to add yourself to the README
- [x ] squash merge the PR with commit message "feat(X): Added 'bar' parameter"


jcubic:
Thanks for the PR there are failing tests I think they are from prettier. You will need to run it locally, since our bot that was fixing the code automatically doesn't work anymore.

jcubic:
It would also be nice if you added unit tests for your changes.

rraab-dev:
Okay I fixed the errors. Now all tests run through on the PR pipeline.

Why I have not added a unit tests is that the unit tests do not run on my setup for some reason. On my machine when I run npm test the result is:
grafik

The strange thing is in the azure pipeline all tests run through, when I fix them all the azure pipeline fails and complains about the fixes and on my windows setup I do not get any prettier errors anymore instead I get this result:
grafik

The error of the failing tests is as follows:
grafik

Do I have to install some local test server first?

My setup:
Windows 10,
Node version: v12.22.12

jcubic:
Ok sorry about that I've thought that unit test were fixed for local run, but it seems not.

PR #1622, opened 2022-08-12
Handle edge cases in walk diff example

MarsOrBust24:
The walk docs example 2 creates difficult to decipher errors when head is undefined or  a document is empty.

sidenote: ` gitdir` is not required... yet it's marked as required. I was able to figure it out, but a complete example would have helped.

```
import { default as isomorphicGitFsClient } from 'node:fs'
import diff from 'diff-lines'
import { fetch, clone, fastForward, currentBranch, walk, TREE, WORKDIR, STAGE } from 'isomorphic-git'
const http = await import('./node_modules/isomorphic-git/http/node/index.js')
const isomorphicGitWorkingTreeDir = '.'

const gitStatus = async () => {
  const ref = 'HEAD'
  const trees = [TREE({ ref }), WORKDIR(), STAGE()]
  const map = async (filepath, [head, workdir]) => {
    return {
      filepath,
      oid: await head?.oid(),
      diff: diff(
        (await head?.content())?.toString('utf8') || '',
        (await workdir?.content())?.toString('utf8') || ''
      )
    }
  }
  const result = await walk({
    fs: isomorphicGitFsClient,
    dir: isomorphicGitWorkingTreeDir,
    trees,
    map
  })
  console.log(result)
}

gitStatus()
```

### DIFF
--- a/src/api/walk.js
+++ b/src/api/walk.js
@@ -177,12 +177,15 @@ import { join } from '../utils/join.js'
  *
  * Example 2: Return the difference between the working directory and the HEAD commit
  * ```js
- * const diff = require('diff-lines')
- * async function map(filepath, [head, workdir]) {
+ * import diff from 'diff-lines'
+ * const map = async (filepath, [head, workdir]) => {
  *   return {
  *     filepath,
- *     oid: await head.oid(),
- *     diff: diff((await head.content()).toString('utf8'), (await workdir.content()).toString('utf8'))
+ *     oid: await head?.oid(),
+ *     diff: diff(
+ *       (await head?.content())?.toString('utf8') || '',
+ *       (await workdir?.content())?.toString('utf8') || ''
+ *     )
  *   }
  * }
  * ```
--- a/src/api/walk.js
+++ b/src/api/walk.js
@@ -177,7 +177,6 @@ import { join } from '../utils/join.js'
  *
  * Example 2: Return the difference between the working directory and the HEAD commit
  * ```js
- * import diff from 'diff-lines'
  * const map = async (filepath, [head, workdir]) => {
  *   return {
  *     filepath,


src/api/walk.js
```diff
@@ -177,12 +177,15 @@ import { join } from '../utils/join.js'
  *
  * Example 2: Return the difference between the working directory and the HEAD commit
  * ```js
- * const diff = require('diff-lines')
- * async function map(filepath, [head, workdir]) {
+ * import diff from 'diff-lines'
```
mojavelinux, line 180: I think we need to stay consistent in the docs whether we're using ESM or CommonJS. Thus for, we have used CommonJS, which is a common denominator. I think we should stick with it here.

MarsOrBust24: CommonJs is a niche tech in the grand scheme of things. of the 36M+ JavaScript developers, less than 20% use Node. Of those who use node, ~50%+ strongly prefer typescript/esm.  Calling a legacy tech on its way out used by less than 5% of the user base, the LCD, is miopic.

I understand where you're coming from in the package maintenance perspective, as it's usually not wise to publish an esm only package at the moment... unless it's a plugin for an esm centric build system like Webpack 5, ESBuild or Vite.

jcubic: If you really think this way please rewrite the whole documentation to use ES Modules. If you don't want to do that then please use whatever is in the project and not introduce a new style or syntax in a random place. If you want to contribute you need to follow the style of the project.

jcubic: This is the first rule of Open Source. Never introduce a new style only follow whatever is in the project. Or your PR will be rejected.

MarsOrBust24: > This is the first rule of Open Source. Never introduce a new style only follow whatever is in the project. Or your PR will be rejected.

This is not a style change this is an English documentation improvement. Documentation is not code and can only be slowly improved upon. If you used tabs for indentation I would have submitted tabs for indentation, despite me being vehemently opposed to them.

Sidenote, the lack of an ESM export for the http module introduces a portability problem, and is a minor hurdle...

...fine I'll get rid of anything that doesn't work with Node v0.10

jcubic: You don't need to be sarcastic. NOTE that by default `import` doesn't work even in NodeJS 18. You need to create package.json file and use `type: module` to be able to use `import` or use some kind of feature flag if there is any.

mojavelinux: The require statement for the diff package is now missing. Can this be corrected?

MarsOrBust24: > 

Yes, in node that is the case. In browsers, deno and bun import works but require doesn't. In node also works if you use the mjs extension. No flags are needed in Node 16 LTS for ES Modules or top level await support.

jcubic: My bad. I didn't noticed that the code is right now broken. Because you use `diff` in the code that is not defined.

PR #1625, opened 2022-08-16
fix(push): force push if no change (oids match)

### LINKED ISSUE #398: should not fail with E.PushRejectedNonFastForward if no new commits to push
Probably happening because the remote head is NOT technically a descendent of the local head. It's the same commit.

It should go ahead and allow the push, because IDK, lets you trigger remote hooks or whatever.

mojavelinux:
resolves #398

Coerce the force flag to true if the oid being pushed matches the oid on the remote.


jcubic:
LGTM

PR #1635, opened 2022-08-31
feat(branch): add object param to specify start point for new branch

### LINKED ISSUE #1634: Add support for start point to branch function
Currently, the `branch` function assumes the new branch should be created starting at HEAD. It would be useful to be able to specify a start point so the new branch can be created independent of where HEAD currently points (meaning you don't have to switch branches first). The branch function should accept the start point via the `object` keyword argument, thus matching the behavior of the `tag` function.

I'm prepare to implement this enhancement. I filed an issue to put forth the proposal and to track the change.

mojavelinux:
resolves #1634

I committed the test repository since the test needs to have multiple different branches configuration.


jcubic:
Looks good

PR #1637, opened 2022-09-02
feat(branch): add force param to overwrite existing branch

### LINKED ISSUE #1636: Add support for force param on branch function
When using the `branch` function, it's not possible to tell isomorphic-git to create a branch (perhaps with a new start point) if it already exists. Instead, it's necessary to check for the branch, delete it, then create it again, which is a lot of extra calls and error handling. Canonical git supports a force flag to reset an existing branch to the specified start point if the branch already exists. isomorphic-git should support this functionality as well through the `force` param on the `branch` function. Adding the `force` param also makes it consistent with the `tag` function.

I'm prepared to implement this enhancement. I filed an issue to put forth the proposal and to track the change.

mojavelinux:
resolves #1636

### DIFF
--- a/__tests__/test-branch.js
+++ b/__tests__/test-branch.js
@@ -49,6 +49,45 @@ describe('branch', () => {
     expect(await listFiles({ fs, dir, gitdir, ref: 'test-branch' })).toEqual([])
   })
 
+  it('branch force', async () => {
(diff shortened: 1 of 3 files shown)


jcubic:
Looks good.

PR #1664, opened 2022-11-10
Report correct expected type in resolveFilePath

HebaruSan:
While investigating an `ObjectTypeError` (I think not this one), I noticed that this spot is checking for `tree` but reporting `blob` as expected in its message. As I understand it they should be the same; now they are.


jcubic:
Thanks for the fix

PR #1665, opened 2022-11-10
Fix filepath option of readTree

HebaruSan:
## Problem

The [docs for `readTree`](https://isomorphic-git.org/docs/en/readTree) say:

filepath | string | Don't return the object with `oid` itself, but resolve `oid` to a tree and then return the tree object at that filepath.
:-- | :-- | :--

However, if you use this option, an `ObjectTypeError` is always thrown:

> `Uncaught ObjectTypeError: Object 678a5b2bab22a1d0ad7c842c1b0dab93d8892384 was anticipated to be a tree but it is a blob.`

Working code snippet:

```javascript
const tree = await git.readTree({fs, dir, cache, oid: commit.commit.tree});
```

Add `filepath` and it throws:

```javascript
const filepath = 'download_counts.json';
const tree = await git.readTree({fs, dir, cache, oid: commit.commit.tree, filepath});
```

## Cause

If `filepath` is set, `readTree` replaces the input oid with the oid of the file blob. But this is then passed to `resolveTree`, which only accepts oids of type `tag`, `commit`, or `tree`:

https://github.com/isomorphic-git/isomorphic-git/blob/0f24b6e56abb4167a9d64c5a825e60d182a3aeda/src/utils/resolveTree.js#L14-L25

So passing a blob oid can't work.

## Changes

Now `readTree` no longer passes a blob oid to `resolveTree`. Instead it calls `resolveBlob` in the `filepath` handler block, so calling code will receive back the blob of the requested filepath.

### DIFF
--- a/src/utils/resolveTree.js
+++ b/src/utils/resolveTree.js
@@ -3,6 +3,7 @@ import { GitAnnotatedTag } from '../models/GitAnnotatedTag.js'
 import { GitCommit } from '../models/GitCommit.js'
 import { GitTree } from '../models/GitTree.js'
 import { _readObject } from '../storage/readObject.js'
+import { resolveBlob } from '../utils/resolveBlob.js'
 
 export async function resolveTree({ fs, cache, gitdir, oid }) {
   // Empty tree - bypass `readObject`
@@ -19,6 +20,9 @@ export async function resolveTree({ fs, cache, gitdir, oid }) {
   if (type === 'commit') {
     oid = GitCommit.from(object).parse().tree
     return resolveTree({ fs, cache, gitdir, oid })
+  }
+  if (type === 'blob') {
+    return resolveBlob({ fs, cache, gitdir, oid })
   }
   if (type !== 'tree') {
     throw new ObjectTypeError(oid, type, 'tree')


HebaruSan:
I don't understand this, but I'll try to obey it...

```
/home/vsts/work/1/s/src/commands/readTree.js
  6:1  error  `../utils/resolveBlob.js` import should occur before import of `../utils/resolveFilepath.js`  import/order
```

HebaruSan:
Hmm. The failing tests seem to be passing **directory** paths in the `filepath` property. That's not a "**file** path". Is this property meant for **file** paths, or not?

jcubic:
Ok, but the tests are failing.

jcubic:
I would blame documentation here not the code. The code is tested and it's working.

The function has `_` in front so it's an internal function, I'm not sure what the public API is doing.

HebaruSan:
The public API is just a passthrough to this code. It just sets a default for params and decorates any exceptions that are thrown:

https://github.com/isomorphic-git/isomorphic-git/blob/0ede4b7ba8e60dca0ea936ead9b5f0cc9acc523a/src/api/readTree.js#L33-L57

HebaruSan:
In case I might have inadvertently misrepresented the docs: If I click through to the docs now, it says "return the tree object", but my copy-paste from earlier says "return the blob object". I'm at a loss to explain this, as I don't see any recent commits for this, and I would have had no reason to edit after pasting. And I think I _remember_ it saying "tree object" as it does now. Did I just miss a commit in another spot?

jcubic:
I'm not sure what to do. I would take a look at broken tests carefully to see what they are testing. Maybe the documentation is wrong. I would trust tests more.

HebaruSan:
I think it's what I said earlier, the tests are passing _directory_ paths in `filepath`. So they resolve to a tree and not a blob. But the property is called `filepath`, which suggests it should work for a file's path, which would resolve to a blob.

I'll try another approach, maybe the tests will be happier if `resolveTree` is able to handle a blob oid instead...

jcubic:
Maybe they're somehow taken from that other PR of yours that fixed the error message.

HebaruSan:
Well, that worked, so I think this approach would address the problem I saw without breaking existing code. But I understand that there may still be a design choice to be made by the team about how `filepath` is meant to work.

jcubic:
I can't merge this I'm not sure what to do. I don't know if this is the right behavior. Now the docs say `return the tree object at that filepath` not `blob`, like in your comment.

Even a copy of the website from says `tree` not `blob`

https://web.archive.org/web/20220523044222/https://isomorphic-git.org/docs/en/readTree

HebaruSan:
Setting the docs aside for a moment, when I see a property called `filepath`, my expectation is that I can set it to the path of a file. I think that's a reasonable expectation, and there are uses cases for it related to inspecting the history of one particular file over time (which is what I'm trying to use isomorphic-git for).

But currently that throws `ObjectTypeError`. I would understand that if the calling code had itself passed the wrong kind of oid somewhere, but it hasn't. So this exception is surprising, and it's what this PR is addressing.

mojavelinux:
Given the fact that the function name is readTree, I don't think it should be returning anything other than a tree. Given this context, the function assumes that filepath is a directory. Asking it to resolve a file would be an error as a file is not a tree. The documentation should emphasize this requirement. I don't think the code should be changed.

HebaruSan:
> In case I might have inadvertently misrepresented the docs: If I click through to the docs now, it says "return the tree object", but my copy-paste from earlier says "return the blob object". I'm at a loss to explain this

Side note resolving this mystery, I think I had multiple tabs open and switched to the one for `readBlob` to copy the `filepath` row from the table. That page says "return the blob object", so that would explain it.

jcubic:
Ok, so I'm closing this PR.

Issue #1495, opened 2022-02-21
Expose lower-level constructs


SamVerschueren:
Hi everyone 👋 !

I was wondering if it would make sense to expose lower-level constructs like the managers and the models. That would give users of `isomorphic-git` more flexibility in building features that are currently not supported by `isomorphic-git`. As an example, the issue I just created, https://github.com/isomorphic-git/isomorphic-git/issues/1494, could probably be built myself if I had access to the `GitIndexManager`.

After that I could always contribute back if we agree on an API that might be helpful for others to use as well. But `isomorphic-git` has such good low-level building blocks, and it's sometimes annoying that it cannot be used directly.

mojavelinux:
We have to be very cautious when exposing internal/utility object-oriented APIs because it increases the API surface area of the library. That can make isomorphic-git harder to refactor in the future without ripping effects. I would be in favor of adding more plumbing functions like we have with resolveRef. But to expose APIs like GitRefManager moves isomorphic-git away from being a functional library and I'm not in favor of it.

SamVerschueren:
Doing a refactor could also be released as a semver major because it's breaking. However, I do see what you are pointing out here which definitely makes sense. I'm trying to find a solution for myself, and maybe also others, to enable more flexibility without constantly taking time from maintainers like you and @jcubic who do this in their spare time.

jcubic:
I personally think that having more objects like `GitRefManager` is not that bad. They can be used for specific low-level handling of the index, cache, etc.

But it's also a good idea to make this a separate project. And later if it will be needed it can be integrated. We can even make it in isomorphic-git org as an experiment.

Isomorphic-git is created using ES Modules so there is no reason not to dig deep ad import internal functions to implement this new library. If something is not exported (internally) we can modify it and export it so it can be used in this new project.

@SamVerschueren you can create a proof of concept of the new low-level API. We need to be very careful about what the API will look like.

I'm thinking about maybe `GitIndex` object that has stuff that is related to git index. And it doesn't have to have all the features. We can add them one by one when needed. When `GitIndex` has everything that is needed, we can integrate it with isomorphic-git and make it a major update.

mojavelinux:
[...]

PR #1669, opened 2022-11-14
feat: expose internal api

### LINKED ISSUE #1071: Internal APIs currently needed for several significant use-cases
First, thanks for the creating this awesome library!

There are a few use-cases which seem to be difficult to implement without access to APIs that are currently designated as "internal".

### Decorated HTTP plugins

The decorator pattern would be a natural choice for those wishing to modify the default http behavior in new ways without completely re-implementing it. However, as far as I can tell, once you replace the default http plugin, there isn't a good way to delegate to it from the decorator.

### Implementing the server-side of pack upload/download

Most of the pieces already exist in isomorphic-git for a client to use the library to implement a git server supporting pushing/pulling changes. I have a prototype which supports git clone over WebRTC which required very little glue code - the main new piece was an 'side-band-64k' protocol implementation based on the commented code already present in isomorphic-git.

Unfortunately, this already required access to several internal parts of isomorphic-git (`uploadPack`, `GitPktLine`, `collect`, `parseUploadPackRequest`, `listObjects`), without which I would have had to duplicate a great deal more code.

hrueger:
fixes #1071 and #1495
It probably needs some documenation (e.g. expect breaking changes not only in new major versions, not officially supported, ...)
I also haven't figured out how to tell typescript that the `internal-apis.d.ts" belongs to the `internal-apis.js` file. The [triple slash directive](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-path-) does not seem to work.

### DIFF
--- a/declaration.tsconfig.json
+++ b/declaration.tsconfig.json
@@ -1,5 +1,5 @@
 {
-  "include": ["index.js", "http/web/index.js", "http/node/index.js"],
+  "include": ["index.js", "http/web/index.js", "http/node/index.js", "internal-apis.js"],
   "exclude": ["node_modules"],
   "compilerOptions": {
     "types": [],
--- a/package.json
+++ b/package.json
@@ -48,6 +48,8 @@
     "index.umd.min.js",
     "index.umd.min.d.ts",
     "index.umd.min.js.map",
+    "internal-apis.js",
+    "internal-apis.cjs",
     "size_report.html"
   ],
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "index.umd.min.js.map",
     "internal-apis.js",
     "internal-apis.cjs",
+    "internal-apis.d.ts",
     "size_report.html"
   ],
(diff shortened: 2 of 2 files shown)


mojavelinux:
While I recognize that there are operations that are needed to accomplish certain tasks with isomorphic-git, I do not support wholesale exposing the entire internal API. It's named that for a reason. It's meant to be internal.

If you can identify specific functionality that should be public, and are willing to define what that public API looks like, I would be in support for that. This is just not the way to go about it.

hrueger:
Understandable. However, unfortunately, I don't have time to pick the bits and pieces I need and create a public api for them.
That means, I'll have to just keep my fork up to date from upstream.
I'll guess, I close this then?

mojavelinux:
As wmhilton said in #1495, you should use a fork if you need to change isomorphic-git for your own needs in a way that isn't compatible with the design of the project. If you have time to contribute a change that would be compatible with the project, then you can do so at any point in the future.

PR #1683, opened 2022-11-25
fix readme: explicit polyfill of Buffer (#1680)

### LINKED ISSUE #1680: docs: Module "buffer" has been externalized for browser compatibility
when i run isomorphic-git in a browser, i get this error

> Module "buffer" has been externalized for browser compatibility. Cannot access "buffer.Buffer" in client code

with vite bundler, this is fixed by adding the `buffer` package

```js
import { Buffer } from 'buffer'
globalThis.Buffer = Buffer
```

https://github.com/vitejs/vite/discussions/2785

lets add this to readme?

https://github.com/isomorphic-git/isomorphic-git#getting-started

> If you're using isomorphic-git in the browser, you'll need something that emulates the fs API.

milahu:
fix #1680

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,12 +86,19 @@ git.clone({ fs, http, dir, url: 'https://github.com/isomorphic-git/lightning-fs'
 If you're using ES module syntax, you can use either the default import for convenience, or named imports to benefit from tree-shaking if you are using a bundler:
 
 ```js
+// npm install buffer
+// explicit polyfill of the nodejs Buffer module, needed for the vite bundler
+import { Buffer } from 'buffer'
+globalThis.Buffer = Buffer
+
 import git from 'isomorphic-git'
 // or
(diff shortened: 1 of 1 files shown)


jcubic:
Please don't add comments to README and don't modify existing examples only add a new section with a header that indicates that this is for Vite. Every other bundler works fine without this hack. So this only obfuscates the examples. This was my main concern when I said that README is not good place for this.

milahu:
> this only obfuscates the examples

this allows to copy-paste the sample code

> Every other bundler works fine without this hack

this code works for all bundlers

jcubic:
> > Every other bundler works fine without this hack

> this code works for all bundlers

Sure but it's redundant, if you want this to be merged please create separated example that is require only for Vite and not need to other bundlers.

PR #1705, opened 2022-12-28
Symlink-fixes

mshanemc:
repro:

1. pass statusMatrix a dir that is a symlink to a folder--you'll see all the files show up as changed
2. pass that folder to `.add` to add all those files
3. ask statusMatrix for the changed files

expected: there are no changes (the were all added)

actual: it shows all the changes

### what was happening
 node fs.lstat was used in a few places.  Callling isDirectory on that will return `false` when it's a symlink.  This change

1. adds the `stat/fstat` methods to the FileSystem to get stats that report about the symlink's target, not the symlink itself.
2. the GitFsWalker will check blob types that are symlinks to determine if they're really directories (`tree`) so that StatusMatrix gets correct answers.

details:
this bug came to us from https://github.com/forcedotcom/cli/issues/1851

## I'm fixing a bug or typo


jcubic:
Can you write tests for this? I think that it should treat symlinks to directories as files. Do your changes make it work like this?

mshanemc:
oh, you're right that git treats symlinks (even for dirs) like files.  Our tooling wants to know what's in the those linked dirs, but that would take isogit away from the git behavior.  

I'm going to close this and try to find another way to track files in symlinked dirs.

mshanemc:
here's the new/improved one, with some tests @jcubic https://github.com/isomorphic-git/isomorphic-git/pull/1706

PR #1706, opened 2022-12-29
fix: avoid recursing symlinked directories

mshanemc:
## I'm fixing a bug or typo

repro:

pass statusMatrix a dir that is a symlink to a folder--you'll see all the files show up as changed

expected: one file (git status treats the symlink like a file)

actual: it shows all the files in the linked dir

what was happening
node fs.lstat was used in a few places. Callling isDirectory on that will return false when it's a symlink. This change

adds the stat methods to the FileSystem to get stats that report about the symlink's target, not the symlink itself.
the GitFsWalker will check blob types that are symlinks to determine if they're really directories (tree) so that StatusMatrix gets correct answers.
details:
this bug came to us from https://github.com/forcedotcom/cli/issues/1851

### DIFF
--- a/__tests__/test-add.js
+++ b/__tests__/test-add.js
@@ -181,6 +181,41 @@ describe('add', () => {
     await add({ fs, dir, filepath: 'c' })
     expect((await listFiles({ fs, dir })).length).toEqual(4)
   })
+  it('symlink to a folder', async () => {
+    // Setup
+    const realFolder = 'realFolder'
+    const { fs, dir } = await makeFixture('test-add')
+    await fs.mkdir(`${dir}/${realFolder}`)
+    await fs.write(`${dir}/${realFolder}/a.txt`, 'a')
+    const symlinkCreated = await fs
+      ._symlink(`${dir}/${realFolder}`, `${dir}/linkToFolder`)
+      .then(() => true)
+      .catch(() => false)
+    // Test
+    await init({ fs, dir })
+    expect((await listFiles({ fs, dir })).length).toEqual(0)
+    if (!symlinkCreated) return
+
+    await add({ fs, dir, filepath: 'linkToFolder' })
+    console.log(await listFiles({ fs, dir }))
+    expect((await listFiles({ fs, dir })).length).toEqual(1)
+
+    const walkResult = await walk({
+      fs,
+      dir,
+      trees: [STAGE()],
+      map: async (filepath, [stage]) =>
+        filepath === 'linkToFolder' && stage ? stage.oid() : undefined,
+    })
+    expect(walkResult.length).toEqual(1)
+    const oid = walkResult[0]
+    const { blob: symlinkTarget } = await readBlob({ fs, dir, oid })
+    let symlinkTargetStr = Buffer.from(symlinkTarget).toString('utf8')
+    if (symlinkTargetStr.startsWith('./')) {
+      symlinkTargetStr = symlinkTargetStr.substr(2)
+    }
+    expect(symlinkTargetStr).toMatch(/.*\/realFolder/)
+  })
   it('folder with .gitignore', async () => {
     // Setup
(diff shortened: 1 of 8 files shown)


__tests__/test-statusMatrix.js
```diff
@@ -394,4 +394,109 @@ describe('statusMatrix', () => {
       ['b/b.txt', 0, 2, 2],
     ])
   })
+
+  describe('statusMatrix with a symlinked directory', () => {
+    const realFolder = 'realFolder'
+    let storedProps = {}
+    const symlinkCreated = false
```
jcubic, line 401: Because of this, the tests actually don't run.

__tests__/test-add.js
```diff
@@ -181,6 +181,41 @@ describe('add', () => {
...
+      ._symlink(`${dir}/${realFolder}`, `${dir}/linkToFolder`)
+      .then(() => true)
+      .catch(() => false)
+    // Test
+    await init({ fs, dir })
+    expect((await listFiles({ fs, dir })).length).toEqual(0)
+    if (!symlinkCreated) return
+
+    await add({ fs, dir, filepath: 'linkToFolder' })
+    console.log(await listFiles({ fs, dir }))
```
jcubic, line 200: Can you remove console.log?

__tests__/test-statusMatrix.js
```diff
@@ -394,4 +394,109 @@ describe('statusMatrix', () => {
...
+    beforeAll(async () => {
+      const { fs, dir, gitdir } = await makeFixture('test-empty')
+      await fs.mkdir(path.join(dir, realFolder))
+      await fs.write(path.join(dir, realFolder, 'a.txt'), 'a')
+      // symlinks can't be created in browser, according to
+      // https://github.com/isomorphic-git/isomorphic-git/blob/main/__tests__/test-add.js#L117      symlinkCreated = await fs
+      const symlinkCreated = await fs
```
jcubic, line 408: This variable is const and it's not used in tests, If think it suppose to be without declaration and outside it should be declared as `let`.

__tests__/test-statusMatrix.js
```diff
@@ -394,4 +394,109 @@ describe('statusMatrix', () => {
       ['b/b.txt', 0, 2, 2],
     ])
   })
+
+  describe('statusMatrix with a symlinked directory', () => {
+    const realFolder = 'realFolder'
+    let storedProps = {}
+    const symlinkCreated = false
+    beforeAll(async () => {
```
jcubic, line 402: Please rewrite the tests and don't rely on sequence, the tests can be run in parallel and it will break the test if one case depends on another case. Use `beforeEach` and prepare new base repo for all tests.

src/models/GitWalkerFs.js
```diff
@@ -46,6 +46,7 @@ export class GitWalkerFs {
   }
 
   async readdir(entry) {
+    if ((await entry.type()) === 'linkTree') return []
```
jcubic, line 49: I think that it would be better to allow follow links in git walker but make it an option, in this case if someone wants old behavior he can still use it. Some users may rely on this feature so make the option `followSymlink: true` default.

jcubic (changes requested):
Two things:
1. Some test tests will not work, they are not executed.
2. Please make an option for gitWalker `followSymlink: true` this way we will not break old behavior and users will decide which option they want. Sorry, just realized this would be breaking change and some users may rely on this feature.

src/models/GitWalkerFs.js
Yesterday17: I think following symlink is never a "defined behavior" in git. Reading dir on such 'files' should work as other files. In fact, I think the signature of `readdir` should be changed to return both file name and file type, making use of Node's `withFileTypes` option instead of calling `stat` again and again.

Issue #763, opened 2019-06-18
support 'pattern' parameter in 'add' command


wmhilton:
`statusMatrix` works great for me. Is there any existing way to use that `pattern` functionality for `add` (i.e. `git add dir/*`), or is the only way to manually add each file?

_Originally posted by @chrisdothtml in https://github.com/isomorphic-git/isomorphic-git/issues/251#issuecomment-471134540_

I think this is a great idea! Adding to the todo list.

pyramation:
fwiw, I had forked `glob` and made a version that works with browser fs. It was a bit hacky, but it works for the sync API of glob. Happy to dig the code out if it's of interest.

wmhilton:
Hopefully I won't need it! But I'll keep it in mind.

PR #1717, opened 2023-01-15
feat(add): add glob expansion to filepath

PouyaEsmaili:
resoloves #763


jcubic:
I don't think that using `filepath` for a pattern is a good idea. Also, the tests are failing.

I would use the `pattern` option and when used the `filepath` should be ignored. @mojavelinux 
What do you think?

mojavelinux:
In canonical git, the primary argument to `git add` (and similar commands) is called a [pathspec](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefpathspecapathspec). And a pathspec does accept wildcard characters.

If we are concerned about a breaking change, we could consider introducing pathspec as an alternative to filepath. That would a) align with canonical git and b) not impact any existing users.

I think if we are going to support wildcards, we should consider using a pattern matching library rather than trying to implement the logic within isomorphic-git. Though, there are benefits of doing it without an additional dependency.

PR #1718, opened 2023-01-15
feat: add GitRemoteFile to support working with local repository as a remote

### LINKED ISSUE #1263: Clone with local-only repositories and branches
The [`clone`](https://isomorphic-git.org/docs/en/clone) command currently requires the `url` parameter to be a URL of the remote repository. However, canonical git allows for local paths to be used when cloning a repository. This becomes an issue when attempting to clone a local-only repository, or using the `ref` parameter to specify a local-only branch for checkout during cloning.

PouyaEsmaili:
resolves #1263


jcubic:
Sorry, but I can't merge this, you've added a big dependency (git-http-backend) that failed the maxSize allowed added by the original author. I need to reject the PR unless you will implement yourself what is needed than Adding such a big library that looks like a git server implementation or if you make this optional package like lightning-fs and not included with the main bundle. Which is important for the browser.

This only for Node there is not need to include the code for the browser.

jcubic:
You will need to rewrite the code to make this:

```
  const http = require('http')
  const backend = require('git-http-backend')
  const spawn = childProcess.spawn
  const zlib = require('zlib')
```

don't be included in the browser. I'm not sure if zlib is already added. And please don't use `require()`, I would expect that this will throw an error from Webpack.

But note that I don't think that running the whole server to clone the repo from the directory is a good idea. You can access the filesystem and .git directory why you need to create an HTTP server. This is kind of silly if you think about it. This is an ugly hack to make this work, not a real solution to the problem.

After giving it a second thought I think that running a HTTP server is out of scope for this project. If you want to clone the repo from the directory you need to read the `.git` directory directly.

jcubic:
May I ask who is `@Nazhixx`?

jcubic:
Why don't you try to run the HTTP server outside of the library? It should work, you don't need to modify isomorphic-git to run the HTTP server to clone the directory.

PR #1719, opened 2023-01-15
feat(clone): add create dir parameter and clone into the created directory

### LINKED ISSUE #625: Create parent directory on clone
Hello,

I'm new to isomorphic-git. I just created a small application that allow me to clone a repo.
Here is my code : 
`await git.clone({
        fs: fs,
        dir: currentPath,
        url: myUrl,
        ref: 'master',
        username: myUsername,
        password: myPassword,
});`

So if my currentPath is /tmp it clones in this directory.
I'm looking for an option to clone in the /tmp/dir-name directory.
For example, if I clone https://github.com/isomorphic-git/ I want to have a isomorphic-git directory created that contain the project. This repo must be created in tmp (/tmp/isomorphic-git).

PouyaEsmaili:
resolves #625

- [ ] add parameter to the function in `src/api/X.js` (and `src/commands/X.js` if necessary)
- [ ] document the parameter in the JSDoc comment above the function


src/commands/clone.js
```diff
@@ -102,13 +112,19 @@ export async function _clone({
...
     // Remove partial local repository, see #1283
     // Ignore any error as we are already failing.
     // The catch is necessary so the original error is not masked.
     await fs
       .rmdir(gitdir, { recursive: true, maxRetries: 10 })
       .catch(() => undefined)
+    if (createDir) {
+      await fs
+        .rmdir(dir, { recursive: true, maxRetries: 10 })
+        .catch(() => undefined)
```
jcubic, line 126: gitdir is inside dir, why do you need to call rmdir on both directories if it uses `recursive: true` option?

src/commands/clone.js
```diff
@@ -102,13 +112,19 @@ export async function _clone({
       remote,
       noCheckout,
     })
+    return { basename: dirname }
```
jcubic, line 115: I would make the API always return basename, but not name of the directory only a full path. But first of all you can change `.git` when using bare repo, so this will break existing API. You need to move `createDir` into `git.clone()` not library's internal function.

jcubic (changes requested):
I think that API could be better and I think that you edited the wrong function. Please read the comments.

jcubic:
Another problem is that there are no tests for new API.

And BTW: you did not create a proper description of a PR. `X` is like a variable you should put there the thing you changed.

Overal this doesn't look very good.

mojavelinux:
If a decision is made to move forward with this idea, I would prefer the option to be named mkdir or mkdirs to align with the common terminology in Node.js fs and libraries that use it. (For example, pino uses mkdirs as the option name).

src/commands/clone.js
PouyaEsmaili: You're right, I can make it one rmdir.

jcubic:
So to sum things up.
* Let's make option `mkdir`.
* The clone method should return `{ path: string }`.
* The code should be in `git.clone()` that allow to change directory.
* I would also add more tests with more options like gitdir used with `mkdir`.

mojavelinux:
> The clone method should return `{ path: string }` always return full path of repo.

If we are going to change the signature, I think perhaps this should be `dir` not `path`. The clone directory is definitely always a path (regardless of whether it's bare or not). We might even want to consider whether to return both `dir` and `gitdir`, for that matter.

jcubic:
Agree `{ dir }` will be much better, because you will be able to merge the result with the settings:

```javascript
let defaults = { ... };

defaults = { ...defaults, ...git.clone({ mkdir: true, ... }) };

git.add({ ...defaults });
```

Issue #1722, opened 2023-01-23
Pushing to private GitHub repository fails


Lotes:
Everything works (clone, branch, add, commit)... even the `onAuth`. Until I `push` to my private repo.

```ts
async function push(dir: string, ref: string) {
  await git.push({ fs, http, onAuth, dir, ref, force: true, remote: 'origin', remoteRef: ref });
}

const pat = () => env.GITHUB_PAT;
const owner = "Lotes";
const ref = "lotes/some/branch";
const onAuth: AuthCallback = () => {
  return {
    username: owner,
    password: pat(),
  };
};
```

I get the following errors:

```
TypeError: Cannot read properties of undefined (reading 'length')
    at StreamReader.read (...\node_modules\isomorphic-git\index.cjs:2340:39)
    ...
ParseError: Expected "unpack ok" or "unpack [error message]" but received "".
    at parseReceivePackResponse (...\node_modules\isomorphic-git\index.cjs:11616:11)
```

I tried out with fine-grained and classic PAT. Any ideas what is going wrong?
I am using isomorphic-git version `1.21.0`.

jcubic:
Do you use an access token? Did you add proper permissions? Can you push from canonical git with the same authentication method?

Lotes:
Yes, thanks for replying :)... it is a personal access token (PAT) from GitHub.

I tried two PATs:
1. a fine-grained one, restricted to the private repo and with option "contents" on READ/WRITE access
2. a classic one with the OAuth scopes "user" and "repo"

Both tokens work with `git push https://Lotes:my_pat@github.com/Lotes/private_repo`.

jcubic:
Then I have no idea, this project is contributor driven I'm a maintainer but I rarely write any code myself. I'm also not that familiar with the code base and git internals. I also don't have much time. I mostly do code reviews and merge Pull Requests. I become a maintainer so the project doesn't die because the original author left.

So if you want this to be fixed you will need to debug this yourself. Any contribution and fixes you can make are welcome.

Lotes:
OK, thanks for your time. I think I can debug it, at least I can try.

Lotes:
Here are my findings:
- `git-upload-pack` is fetched by the push command
- the response is passed to  `GitSideBand.demux`
- whose resulting packfile is passed into `parseReceivePackResponse`
- but packfile is empty

jcubic:
And the push was not successful? There are no data on GitHub?
[...]

PR #1728, opened 2023-02-03
This addresses changes to auth API #1722

konsumer:
I removed all the header stuff (since it's not in the `push` function anymore) and added notes on adding it to the URL or using `onAuth`

Related: #1722

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -6,13 +6,30 @@ sidebar_label: Authentication
 Authentication is normally required for [`push`](./push.html)ing to a git repository.
 It may also be required to [`clone`](./clone.html) or [`fetch`](./fetch.html) from a private repository.
 Git does all its authentication using HTTPS Basic Authentication.
-Usually this is straightforward: just specify `username` and `password`.
+Usually this is straightforward: just specify `username` and `password` in the URL. In a browser, you can do this:
 
 ```js
+// this will just build a url like https://user:password@github.com/isomorphic-git/isomorphic-git
+// it escapes all the non-URL characters for you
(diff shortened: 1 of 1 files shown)


mojavelinux:
I'm good with this change. The permutations of oauth options are already covered on the onAuth page and is thus redundant here. So I believe it is wise to remove that matrix, which is difficult to maintain anyway.

Given what was on this page, I honestly think it had just been forgotten about. The shorter we can make it, the better.

konsumer:
Yeh, I removed the matrix because these are no longer options to `push`, as shown in the example. I tried to show quickly "this works for gitlab/github/bitbucket, and you can also ask user/filesystem on fail with `onAuth`" (which I linked to.) It may be less complete/thorough than it was before, but now it's all working examples.

Hopefully that helps.

mojavelinux:
Yep, we're on the same page.

jcubic:
OK if @mojavelinux thinks this it good. I will merge it.

jcubic:
Hi, I think that you edited the wrong file. Your changes are not public the file on the website is located at:

https://github.com/isomorphic-git/isomorphic-git.github.io/edit/source/docs/authentication.md

Which you can see if you click the edit button on:

https://isomorphic-git.org/docs/en/authentication

I'm not sure what is the purpose of the file you've edited.

jmarks-joshua:
Yeah this is still an issue. Also didn't github disable basic https password authentication in August 2021?

So even without 2fa you cannot just use your email address and password and need to use 'personal access tokens', but then they should have an expiry so need to keep them up-to-date somehow.

PR #1731, opened 2023-02-12
feat(GitIndex): Support multiple index stages

### LINKED ISSUE #1726: Any "git ls-files --stage" analogue?
In case of conflict, git stores multiple versions of file in index.

Here's an example:
```
➜  tmpgit git:(master) ✗ git ls-files --stage
100644 e8e448b33545a4ff26011d77ccc46c0b879f50be 1	index.html
100644 da9736cb76038951b7931f826336a23d21ffb03a 2	index.html
100644 f4952f928d1644e8bfcccc5c96cee07712d7dbaa 3	index.html
100644 ec0966979aa7da1907f1c205b87c173050d7a223 0	script.js
```

How can I get similar information from isomorphic-git?

### LINKED ISSUE #1727: Update GitIndex to include different stages of a path
In case of a conflict, the "original" git stores multiple versions of a path in the index, under different "stage" numbers.

Here's an example of `ls-files` in a repo with a conflict on `index.html`:
```
➜  tmpgit git:(master) ✗ git ls-files --stage
100644 e8e448b33545a4ff26011d77ccc46c0b879f50be 1	index.html
100644 da9736cb76038951b7931f826336a23d21ffb03a 2	index.html
100644 f4952f928d1644e8bfcccc5c96cee07712d7dbaa 3	index.html
100644 ec0966979aa7da1907f1c205b87c173050d7a223 0	script.js
```

`GitIndex` uses `_entries` as `Map(path => data)`, so it's incapable to represent this structure.

The suggestion is to change it. Make it an array instead OR store different stages in `data`.

araknast:
This code implements support for storing files at multiple stages in the git index, e.g. after a merge with unresolved conflicts. 

Since `GitIndex` uses the `Map()` type internally, it is not possible to store different versions of the same file directly without breaking the api. Instead, here we introduce a `stages` property for each index entry, an array whose nth entry represents the nth stage of the file. 

We also introduce an `unmergedPaths` property on `GitIndex` itself, which contains all the index paths that exists at multiple stages. Therefore it is possible to test if the index is unmerged simply by testing `unmergedPaths.length`.

We also introduce the `entriesFlat` getter, which returns a list of all entry objects at all stages in the index. This is in order to avoid breaking the interface of `entries`, which only returns the first stage for every entry in the index.

Finally, we also introduce a new `UnmergedPathsError` if a command which requires a merged index is run with an unmerged one. As far as I know, the only commands which need to throw this are `git.merge` and `git.commit`. 

This should fix #1727 and fix #1726 (use entriesFlat to get unmerged entries).

This commit does not introduce the code to allow `git.merge()` to write multiple index stages in the event of a conflict, that code is [here](https://github.com/isomorphic-git/isomorphic-git/commit/7687e386d97ac86dfbd5a800c379132d7ce556f8) with the implementation of `abortMerge` so users do not accidentally end up in an index state that is difficult to get out of.

If this commit is able to be merged I will then put `abortMerge` up for review as well, as it depends on and multi stage index implementation.


jcubic:
Thanks for your contribution I will check it more carefully when I will get back home from a short trip.

araknast:
Hello, any updates on this?

jcubic:
Sorry, I was busy lately, and I was not able to find the time to dive into the PR.

__tests__/test-merge.js
```diff
@@ -551,6 +578,7 @@ describe('merge', () => {
     // Test
     const testFile = `${gitdir}/o.conflict.example`
     const outFile = `${dir}/o.txt`
+    const cache = {}
```
jcubic, line 581: What is the use of this constant here?

src/models/GitIndex.js
```diff
@@ -35,11 +35,28 @@ export class GitIndex {
...
+      entry.stages = [entry]
+      this._entries.set(entry.path, entry)
+      this._unmergedPaths.delete(entry.path)
+    } else {
+      let existingEntry = this._entries.get(entry.path)
+      if (!existingEntry) {
+        this._entries.set(entry.path, entry)
+        existingEntry = entry
+      }
+      existingEntry.stages[entry.flags.stage] = entry
```
jcubic, line 55: This looks odd, if entry doesn't exist you will have

```javascript
entry.stages[entry.flag.stage] = entry;
```

is this correct?

src/models/GitIndex.js
```diff
@@ -198,31 +249,44 @@ export class GitIndex {
...
-        writer.writeUInt32BE(stat.gid)
-        writer.writeUInt32BE(stat.size)
-        writer.write(entry.oid, 20, 'hex')
-        writer.writeUInt16BE(renderCacheEntryFlags(entry))
-        writer.write(entry.path, bpath.length, 'utf8')
-        return written
-      })
-    )
+    writer.writeUInt32BE(this.entriesFlat.length)
+    const entryToBuffer = async entry => {
```
jcubic, line 253: This function seems to be pure, can you move it as a method to the class?

jcubic (changes requested):
Two comments, one question and refactor request.

src/models/GitIndex.js
araknast: If you mean that the entry has a circular reference in thestages array then yes, this is intended behavior.

src/models/GitIndex.js
araknast: done

src/models/GitIndex.js
jcubic: Can you maybe write a unit test to check this case? Will this happen often? I don't want to end up with stack overflow when a recursive function will be executed on this cycle.

araknast: This occurs whenever a new stage 0 entry is added so yes, it does occur very often. This also occurs whenever the index is read from disk, so i believe it is already covered by the `'read existing unmerged index'` test case. 

The way I have currently written this, entries on every level have a reference to themselves in the `stages` array. Currently the circular references do not seem to be causing any issues as there is no function in the code that addresses them recursively, but I do understand this can cause issues e.g. with `JSON.stringify` which will throw an error. (This can be fixed by using a JSON library that supports circular references, like CircularJS)

The alternative approach would be to remove the circular reference, and have the entry itself represent the lowest stage of the respective entry. This would remove the circular reference, but at the cost of removing assurance that the nth entry in the `stages` array is the nth stage of the entry, which would require extra logic to handle.

jcubic: ok, so maybe write in the documentation about these potential issues, working for users to be careful with recursive code when reading stages.

araknast: Ok, which section of the documentation should this go into? There doesn't seem to be a section of the documentation for internal APIs, should I create one?

jcubic: Oh if there are no docs for GitIndex I think that we can leave the documentation out.

PR #1744, opened 2023-03-13
feat(abortMerge): Introduce 'abortMerge' command 

araknast:
This code introduces new functionally to `git.merge`. If `git.merge` is called with the parameter `abortOnConflict: false`, then in the event of a merge conflict, in addition to adding conflict markers to the problem files, the index will also now record the 3 versions of the conflicted file at different stages, in the same way as canonical git. 

Since certain operations should not be allowed on an unmerged index, we also introduce the `UnmergedPathsError`, which is thrown whenever a command such as `git.commit` is called when the index contains unmerged paths.

A new command `abortMerge` is also introduced, which resets the index and worktree in a way which is *nearly* (see comment in `abortMerge.js`) identical to the behavior of `git reset --merge` and the common behavior of `git merge --abort`. The behavior is based on the documentation of `git reset --merge`, i.e:

"Resets the index and updates the files in the working tree that are different between <commit> and HEAD, but keeps those which are different between the index and working tree (i.e. which have changes which have not been added). If a file that is different between <commit> and the index has unstaged changes, reset is aborted."

In the case that the index and worktree cannot be reset, an `IndexResetError` is thrown and the reset is aborted.
 

## I'm adding a new command:
- [x] add as a new file in `src/api` (and `src/commands` if necessary)
- [x] update `__tests__/test-exports.js`
- [x] document the command with a JSDoc comment
- [x] add page to the Docs Sidebar `website/sidebars.json`
- [x] add page to the v1 Docs Sidebar `website/versioned_sidebars/version-1.x-sidebars.json`
- [ ] squash merge the PR with commit message "feat: Added 'X' command"


.gitignore
```diff
@@ -75,7 +75,12 @@ node_modules
 # Webstorm development environment
 .idea/
 
+# Vim swap files
+*.swp
+*.swo
```
jcubic, line 80: ONe note you can add globally those files to global gitignore you don't need to commit that to the repo when you use vim:

See SO: [Global Git ignore](https://stackoverflow.com/q/7335420/387194)

jcubic (approved):
Looks good, one thing is about gitignore you can configure your own global gitingore where you put vim swap files.

But I think it's ok to merge.

Issue #1751, opened 2023-04-14
'No such file or directory, stat 'somepath/.git/config'' when cloning a repo


CluEleSsUK:
Using ts-node@10.9.1 and node 19.4.0, I ran the following snippet:
```typescript
const repo = "https://some.repo/here"
const dir = path.join(process.cwd(), "./blah")
await fs.mkdir(dir)
await git.clone({fs, http, dir, url: repo, singleBranch: true, depth: 1})
```
and get the error 
```[Error: ENOENT: no such file or directory, stat 'some/path/to/blah/.git/config'] { errno: -2, code: 'ENOENT', syscall: 'stat', path: 'blah/.git/config' }```

I wouldn't expect `./blah/.git/config` to exist yet as I'm cloning the repo for the first time.
I also tried running `git.init()` before it but it also threw the same error.
The repo in question was public and can be cloned fine using the vanilla git CLI

**Comment by jcubic** [comment]
What is `http` in your code? Maybe the library throws a wrong error. Also, can you run this with node? You don't have any types with this example.

**Comment by CluEleSsUK** [comment]
my imports are as follows:
```javascript
import path from "path"
import fs from "fs/promises"
import * as http from "isomorphic-git/http/node"
import * as git from "isomorphic-git"
```

jcubic:
did you tried:
```javascript
import fs from "fs"
```
like in the documentation (npm readme)?

CluEleSsUK:
Ah hmm, yes using `fs` without the promises seems to work - I would have expected the promise version to work too ;o

CluEleSsUK:
Understood - thanks for your swift help!!

slyduda:
I am running into the same issue on `git.init` on a WebWorker based implementation with the following code:

```js
importScripts(
    "https://unpkg.com/@isomorphic-git/lightning-fs",
    "https://unpkg.com/isomorphic-git@beta",
    "https://unpkg.com/isomorphic-git@beta/http/web/index.umd.js",
);

const fs = new LightningFS('project');
const pfs = fs.promises;

await pfs.writeFile('/path', 'Something')

const init = await git.init({ fs: pfs, dir: '/' })
const fileList = await git.listFiles({ fs: pfs, dir: '/' })

```

The error code again is as follows: Error: `ENOENT: /.git/config`

I tried following the WebWorker example in the docs and made promise modifications. Am I missing an import that is causing this issue?

jcubic:
[...]

PR #1757, opened 2023-05-02
Refactor `FileSystem` and add support for `fs/promises`

araknast:
These commits allow isomorphic-git to work natively with node's `fs/promises` module as well as any filesystems which only provide promise-based functions, and do not expose a `.promises` object. Fixes the issue seen in #1751.

When `fs.promises` does not exist, we now check to see if an arbitrary command (in this case`fs.readFile`) returns a promise. If it does, the filesystem is identified as a promise-based filesystem and treated accordingly. Otherwise, it is treated as a callback-based filesystem and wrapped using `pify` as before.

These changes result from of a refactor of the code, which hopefully should make extending `FileSystem` easier in the future.

Additionally, I've updated the website docs to match with the ones generated by `jsdoc`.

### DIFF
--- a/src/models/FileSystem.js
+++ b/src/models/FileSystem.js
@@ -4,6 +4,57 @@ import { compareStrings } from '../utils/compareStrings.js'
 import { dirname } from '../utils/dirname.js'
 import { rmRecursive } from '../utils/rmRecursive.js'
 
+function isPromiseFs(fs) {
+  const test = targetFs => {
+    try {
+      // If readFile returns a promise then we can probably assume the other
+      // commands do as well
+      return targetFs.readFile().catch(e => e)
+    } catch (e) {
+      return e
+    }
+  }
+  return test(fs).constructor.name === 'Promise'
+}
+
+// List of commands all filesystems are expected to provide. `rm` is not
+// included since it may not exist and must be handled as a special case
+const commands = [
+  'readFile',
+  'writeFile',
+  'mkdir',
+  'rmdir',
+  'unlink',
+  'stat',
+  'lstat',
+  'readdir',
+  'readlink',
+  'symlink',
+]
+
+function bindFs(target, fs) {
+  if (isPromiseFs(fs)) {
+    for (const command of commands) {
+      target[`_${command}`] = fs[command].bind(fs)
+    }
+  } else {
+    for (const command of commands) {
+      target[`_${command}`] = pify(fs[command].bind(fs))
+    }
+  }
+
+  // Handle the special case of `rm`
+  if (isPromiseFs(fs)) {
+    if (fs.rm) target._rm = fs.rm.bind(fs)
+    else if (fs.rmdir.length > 1) target._rm = fs.rmdir.bind(fs)
+    else target._rm = rmRecursive.bind(null, target)
+  } else {
+    if (fs.rm) target._rm = pify(fs.rm.bind(fs))
+    else if (fs.rmdir.length > 2) target._rm = pify(fs.rmdir.bind(fs))
+    else target._rm = rmRecursive.bind(null, target)
+  }
+}
+
 /**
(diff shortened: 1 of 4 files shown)


src/models/FileSystem.js
```diff
@@ -4,6 +4,57 @@ import { compareStrings } from '../utils/compareStrings.js'
...
+function bindFs(target, fs) {
+  if (isPromiseFs(fs)) {
+    for (const command of commands) {
+      target[`_${command}`] = fs[command].bind(fs)
+    }
+  } else {
+    for (const command of commands) {
+      target[`_${command}`] = pify(fs[command].bind(fs))
+    }
+  }
```
jcubic, line 44: You don't check if `fs.promises` is provided and always use `pify` that is a fallback for callback-based fs without promises interface in old code.

jcubic (changes requested):

jcubic:
Can you also check if the library works with LightingFS? I'm not sure if this is checked by unit tests.

araknast:
> Can you also check if the library works with LightingFS? I'm not sure if this is checked by unit tests.

Should be covered by the `karma` browser tests for Firefox and Chrome as long as the `ENABLE_LIGHTNING_FS` env variable is set, according to the code in [`FixtureFS.js`](https://github.com/inlang/isomorphic-git/blob/e3fab7be6eea245146bede67cd41f50152708580/__tests__/__helpers__/FixtureFS.js#L13-L20)

src/models/FileSystem.js
jcubic: But you can use the `fs` system that has `fs.promises`, but the user passes `fs`, so you use pify to create promises for no reason if there is already `fs.promises`. There are 3 cases here not two.

araknast: I believe each case is handled correctly no? The execution paths should be:

1. `fs` is passed and `fs.promises` exists
    - `if (promises && promises.enumerable) ` block is triggered
    - `bindfs(fs.promises)` is called
    - `fs.promises` is a promise fs, so `isPromiseFs` evaluates to `true`
    - **commands are not wrapped using `pify`**
2. `fs` is passed and `fs.promises` does not exist, but `fs` is a promise based fs
    - `if (promises && promises.enumerable) ` evaluates to `false`
    - `else` block is triggered
    - `bindfs(fs)` is called
    - `if (isPromiseFs(fs)) ` evaluates to `true`
    - **commands are not wrapped using `pify`**
3. `fs` is passed and `fs.promises` does not exist, and `fs` is not a promise based fs
    -  `if (promises && promises.enumerable) ` evaluates to `false`
    - `else` block is triggered
    - `bindfs(fs)` is called
    - `if(isPromiseFs(fs))` evaluates to `false`
    - **commands are wrapped using `pify`**

If this is incorrect, could you explain the execution path where this does not work, or explain the error in the execution paths I've laid out?

jcubic (changes requested):

src/models/FileSystem.js
araknast: > Where is fs.promises.rmdir.bind code?

If `fs.promises` exists, `bindFs(this, fs.promises)` is called, so the `fs` argument to `bindFs` evaluates to `fs.promises`.

Then `if (isPromiseFs(fs))` evaluates to `true`  because `fs.promises`, has been assigned to `fs`, and `fs.promises` is a promise fs.

Since `if (isPromiseFs(fs))` evaluates to `true` then we call `target[`_${command}`] = fs[command].bind(fs)` in a loop.

 `fs.promises` has been assigned to `fs` in the function call, so this statement evaluates to `target[`_${command}`] = fs.promises[command].bind(fs.promises)`

jcubic: Ok, now I understand. Looks good.

jcubic (approved):

PR #1759, opened 2023-05-10
replace "unset" string value of gitattributes with 'unset' syntax

Byron:
The string "unset" doesn't actually unset them, but sets them to an invalid value.
With this change the attributes will be unset which should have the same effect as before.

----

Please feel free to close this PR if `eol=unset` is entirely intentional. The reason I am proposing this change is merely to help my `gitoxide` validation tool to succeed, and it arrives at 'the wrong' conclusion as it only interprets strings when running `git` on the command-line and "unset" always has to mean "an unset attribute".

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,7 +1,7 @@
 * eol=lf
-__tests__/__fixtures__/** binary eol=unset linguist-vendored
-__tests__/__snapshots__/** binary eol=unset
+__tests__/__fixtures__/** binary -eol linguist-vendored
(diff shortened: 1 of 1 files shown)


mojavelinux:
Given the first line in .gitattributes (`* eol=lf`), I'm pretty certain this is the right change.

Byron:
I'd think that if CI runs on both Linux and Windows, it's at least clear that the change doesn't break anything. I don't know what kind of runners are used though. Of course this could be put to the test by removing all the 'unset' lines.

PR #1772, opened 2023-06-02
fix #1650 - Added clarification in the api documentation.

### LINKED ISSUE #1650: `listBranches` return nothing on init project
Scenario:
1. Create a project
2. Make a `git.init` command
3. get all local branch with
```
git.listBranches({
  fs,
  dir: 'CURRENT_PROJECT_DIR',
}),
```

I expect to have `['master']` as branche list, but I receive `[]`.
And when I use this command: 
```
  git.currentBranch({
    fs,
  dir: 'CURRENT_PROJECT_DIR',
    fullname: false,
  });
```
I receive `'master'`

So I think, there I think there is a bug :)

Using with browser and browserfs with import statement in my vue 3 composable: 
```
import git from 'isomorphic-git';
import http from 'isomorphic-git/http/web';
```

dead-end:
I thought about the issue and I think returning an empty list when there is no initial commit is correct but unexpected, so I added a note in the api documentation.

### DIFF
--- a/src/api/listBranches.js
+++ b/src/api/listBranches.js
@@ -15,6 +15,9 @@ import { join } from '../utils/join.js'
  * If you want an up-to-date list, first do a `fetch` to that remote.
  * (Which branch you fetch doesn't matter - the list of branches available on the remote is updated during the fetch handshake.)
  *
+ * Also note, that a branch is a reference to a commit. If you initialize a new repository it has no commits, so the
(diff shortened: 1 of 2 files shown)


jcubic:
I'm not sure if the fix will be released. I've changed the commit message to `docs:` prefix because I've found PR like this from the original author.

PR #1778, opened 2023-06-18
fix: use lstat bigint=true for local environments to fix #1245

### LINKED ISSUE #1245: Git status reports "unmodified" when an integer in a file is changed
I'm running into an issue whereby if I change a single integer in a file, `git.status()` reports `"unmodified"` when it should be `"*modified"`. Here is a minimal repro:

<https://github.com/alex996/git-int-test/blob/master/int.json>

```json
0
```

`index.js`

```js
const { clone, status } = require('isomorphic-git')
const http = require('isomorphic-git/http/node')
const rimraf = require('rimraf')
const fs = require('fs')
const { strictEqual } = require('assert')

main().catch(console.error)

async function main () {
  const dir = './demo'
  const pfs = fs.promises

  rimraf.sync(dir)

  await pfs.mkdir(dir)

  await clone({
    fs, http, dir, url: 'https://github.com/alex996/git-int-test.git', ref: 'master',
  });

  let int = JSON.parse(await pfs.readFile(`${dir}/int.json`, 'utf-8')); // 0

  int++; // 1

  await pfs.writeFile(`${dir}/int.json`, JSON.stringify(int, null, 2) + '\n', 'utf-8')

  strictEqual(await status({ fs, dir, filepath: 'int.json' }), "*modified")
}
```

When I run `node index.js`, I get

```sh
+ actual - expected

+ 'unmodified'
- '*modified'
    at main (/home/alex/Contract/caravan/git-memdb-demo/index.js:27:3) {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: 'unmodified',
  expected: '*modified',
  operator: 'strictEqual'
}
```

Here are the deps in `package.json`

```json
"dependencies": {
  "isomorphic-git": "^1.7.8",
  "rimraf": "^3.0.2"
}
```

Also here's my env:

```sh
$ node -v
v14.13.1

$ npm -v
6.14.8

$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.1 LTS
Release:	20.04
Codename:	focal
```

Strangely enough, double-digit integers are registered correctly: `int = 10;`. I didn't notice this with strings, only integers so far.

Initially, I ran into this with objects that had integer props, something like: `{ "int": 0 }`, hence the use of the JSON API.

Any idea where the culprit might be? Thanks

jayree:
## I'm fixing a bug or typo

- [x] Browser: what should be returned by `SecondsNanoseconds` if `nanoseconds === undefined`

this fix uses `lstat` with `bigint: true` to be able to compare `mtimeNs` with `mtime nanosecond` from git index


dead-end:
Hi, I just saw that you commit a fix for #1245. I tried this a few weeks ago and we had to revert my fix. The issue was that there were small differences in the timestamp in the index file and what lstat returned although the content did not changed. The result was that the sha1 had to be computed for files that did not change. For a large repo the time to get the statusmatix took minutes.

I do not know where the small differences in the timestamp came from, but I found that the whole issue is known as [race git](https://git-scm.com/docs/racy-git/en), which is described in the article.

Are you sure that your fix works?

jcubic:
The tests are not passing, it seems that the pipeline run for way too long as it uses it. I'll try to restart the pipeline if it still fails it's probably an issue with the PR.

jayree:
> I do not know where the small differences in the timestamp came from, 

@dead-end these small differences in nanoseconds occur because nanoseconds cannot be calculated from milliseconds `const nanoseconds = (milliseconds - seconds * 1000) * 1000000`. I'm using `mtimeNs` and `ctimeNs` (https://nodejs.org/api/fs.html#statsctimens) but this fs option doesn't seem to be compatbile with the browser @jcubic

jcubic:
Can you write a unit test for the use case in https://github.com/isomorphic-git/isomorphic-git/issues/1245? When the author is writing the number to the file. If it works we will be able to close that issue.

__tests__/test-status.js
```diff
@@ -78,4 +78,31 @@ describe('status', () => {
...
+
+  /**
+   * The unit test checks the fix for:
+   *
+   * https://github.com/isomorphic-git/isomorphic-git/issues/1245
+   *
+   * Without the fix, the second write is not detected and the status is 'added'
+   * instead of '*added'
+   */
+  ;(process.browser ? xit : it)(
```
jayree, line 90: > Can you write a unit test for the use case in #1245? When the author is writing the number to the file. If it works we will be able to close that issue.

@jcubic I've reused @dead-end's test

__tests__/test-status.js
```diff
@@ -78,4 +78,31 @@ describe('status', () => {
...
+
+      await Promise.all([
+        await fs.write(path.join(dir, file), 'Hi'),
+        await add({ fs, dir, filepath: file }),
+        await fs.write(path.join(dir, file), 'Ho'),
+      ])
+
+      // Test
+      const a = await status({ fs, dir, filepath: file })
+      expect(a).toEqual('*added')
```
jayree, line 105: > Can you write a unit test for the use case in #1245? When the author is writing the number to the file. If it works we will be able to close that issue.

@jcubic I've reused @dead-end's test

jayree:
@jcubic before merging, we need to evaluate if the changes `SecondsNanoseconds` break something else. `normalizeStats` is also used in other cases!

jayree:
@jcubic I think I finally found a reliable way to support both scenarios (local and browser).

It looks like millisecond precision is enough for the browser scenario.

src/models/FileSystem.js
```diff
@@ -222,6 +222,12 @@ export class FileSystem {
   async lstat(filename) {
     try {
       const stats = await this._lstat(filename)
+      // For non-browser (local) scenarios regular git 'add' command might be used to write the index. Therefore we need to have the exact nanoseconds (see. normalizeStats.js SecondsNanoseconds ).
```
jayree, line 225: > So I think the issue occurs if users use their repository with git and isomorphic-git. If you stage a file with git and then call statusMatrix with isomorphic-git then you compare the git lstat with the isomorphic-git lstat and there can be differences. And if you have differences then you throw away the cache and get a performance issue.

@dead-end thats right, and is mentioned here. if you use native git the timestamp (nanoseconds) is more accurate than the timestamp ('faked' nanoseconds by using ms * 1000000) in isomorphic-git. (this is also the reason why the git index is broken after isomorphic-git tries to rebuild it, becuase it uses the 'faked' nanoseconds) 

Thats why we need to differentiate between browser and local. With `bigint: true` isomorphic-git is able to create the same timestamp (nanoseconds) as native git does.

__tests__/test-GitIndex.js
```diff
@@ -117,4 +117,15 @@ describe('GitIndex', () => {
       expect(index.entriesMap.get('c').stages.length).toBe(1)
     })
   })
+
+  it('ensure nanoseconds are always present', async () => {
```
jayree, line 121: > I think it would be a good idea to write a test that addresses this issue. We can create a repository with git that contains a file and add it to the test repos and then compare the isomorphic-git lstat values with the values in the index file.

@dead-end We can try to enhance this one but I don't know if we can hardcode timestamps to test them the way you want.

dead-end: If you create a repo with git, then this should have an index file with the timestamps. Call `GitIndex.fromBuffer` to get the index content and then call lstat for each entry and compare the values. I think this should be sufficient.

src/models/FileSystem.js
dead-end: @jayree You have written: 

_(this is also the reason why the git index is broken after isomorphic-git tries to rebuild it, becuase it uses the 'faked' nanoseconds)_

Can you explain how the faked nanos lead to the broken index file? We have a lot of issues with this and I would like to understand it.

jcubic:
I'm wondering if the original issue code is correct. It uses direct git add without await like the unit test. I need to try to recreate the issue but with `await`.

__tests__/test-GitIndex.js
jayree: @dead-end tried it with this approach: https://github.com/isomorphic-git/isomorphic-git/commit/e6c277ace38d134368566a4cf7c11952380dd94f but as 'nano.txt' is checked out on the build server, the timestamp changes, so the test will always fail.

dead-end: @jayree Sorry, I did not think of that. The only possibilities I can think of are calling git from node or put the nano.txt in a zipfile and unzip the file during the test run (this should preserve the timestamps). I do not know if this is possible.
@jcubic

src/models/FileSystem.js
jayree: @dead-end with the old behaviour nanoseconds were calculated based on `ms`. 

From a file with `mtimeMs: 1683216016493.005` 
you'll get `1683216016s` and `493.0048828125ms` --> `493004882.8125ns` 
Due to the numeric data types, floating point values are used, which is not the case in the BigInt data structure.
If you use the `mtimeNs: 1683216016493004997n` (bigint) value
you'll get `1683216016s` and `493004997ns`. These are the vaules git uses in the index. 

The old logic fixing the #1245 compared `493004882.8125ns` with `493004997ns` which caused Isomorphic-git to think the index is broken telling isomorphic-git to rebuild the index, but with `1683216016s` `493004882.8125ns` wich doesn't match the real `ns` timestamp of the file.

jayree:
@jcubic The same question should `isomorphic-git` and native Git be able to coexist? Should `isomorphic-git` be able to read/write to an git index created with `git add` and vice versa?

jcubic:
@jayree yes, I think so, isomorphic-git is stateless everything happens inside the filesystem (in .git directory).

PR #1784, opened 2023-06-24
fix(renameBranch): prevent invalid HEAD state, when renaming current branch

### LINKED ISSUE #1783: Bug: renaming the current branch doesn't update HEAD
Renaming the current branch leaves HEAD untouched by default, which leads to an invalid state of HEAD pointing to a missing ref. 
There is a `checkout` option which if `true`, updates HEAD. But even when `checkout` is not `true`, it doesn't make sense not to update HEAD, if the branch being renamed is the current branch.

I've tested the issue in Browser, not Node.

alirezamirian:
closes #1783

## I'm fixing a bug or typo


jcubic:
The first run of the test failed but after a rerun got an error from SauceLabs, it seems they have a problem with MaoOS and iOS.

https://status.saucelabs.com/

We need to wait until they fix it and re-run the tests.

jcubic:
Now finally tests on Safari work, but it fails on your unit test:

```
05 07 2023 17:48:04.331:ERROR [Edge 79.0.309.65 (Windows 10) | renameBranch | branch already exists]: Expected false to be true.
    at <Jasmine>
    at UserContext.<anonymous> (http://localhost:9876/base/__tests__/index.webpack.js?31bbe4d0e5d12383b5d39fdcb72b281bd1717323:13662:56)
```

alirezamirian:
It failed on the first test case, not the one added in this PR. However, the whole test suite passes on my local env.

alirezamirian:
Looked at the failed test case. I can't see how its failure could be related to the change in this PR, since the expected exception is thrown before the newly added lines. Any thoughts?

jcubic:
Maybe you modify the fixture that is used for two different tests.

jcubic:
Maybe try to restore the original state of the fixture, rename the branch back, and test if it's renamed, I'm not sure how they are used by the tests.

With this type of issue (when it works locally) you need to push the changes and see how CI/CD runs.

jcubic:
It seems that the tests run in sequence:
first, it renames branch `test-branch` to `other-branch` and next test it renames `other-branch` into `test-branch-2`. Your test is last. I have no idea how it can fail.

alirezamirian:
The fixture is not touched as far as I can see. And my git status is clean.
What I'm unsure about is the import to `currentBranch.js`:
```js
import { _currentBranch } from './currentBranch.js'
```
I see it's imported from '../commands/currentBranch.js'. No idea how it could possibly cause this particular test case to fail like this, but I'm gonna change it to be similar to other commands, and see if it makes any difference.

jcubic:
It seems that the first test fails and returns `false` instead of `true`. So it throws a different exception. You can modify the code and console.log the error.

alirezamirian:
Now the same test case failed on a different browser. I think it may be about tests leaving side effects on the fixture. 
In the added test case, master is renamed to other-branch, which is probably the reason the first test case fails, because the first test case tries to rename master to something else. Will look into changing the test so that it won't affect the other tests, or make it not have side-effects on the fixture.

alirezamirian:
Changed the new test case to rename the current branch back to master, to keep the fixture state unchanged. It seems to have fixed the issue.

jcubic:
Is it ok to merge? You deleted your comment.

jcubic:
I have no idea looks like a GitHub issue. But I'm also not sure how the CI/CD works, with the whole Azure thing.

alirezamirian:
We should be good to merge now.

Issue #1789, opened 2023-07-10
Objects present in a pack-file as well as loose break expandOid


willstott101:
We're using isomorphic git server-side to expose a GraphQL interface to some repos. We also serve these repositories from the same filesystem using `node-git-server`.

We've ended up with a repository which has an object both loose, as well as in a pack file. This object's contents is correct in both cases, it's just present twice. I don't actually know if that's something to do with upload-pack on the server-side creating and saving packs for clients which it then doesn't actually use as primary storage, or if something else has gone wrong.

The error we see is `"Found multiple oids matching \"c6f444fb83d643dbc82bedbc02e2cf0eccf204d4\" (c6f444fb83d643dbc82bedbc02e2cf0eccf204d4, c6f444fb83d643dbc82bedbc02e2cf0eccf204d4). Use a longer abbreviation length to disambiguate them."`

I have a test and a "fix" but I'm not sure I know enough to deem it definitely a good solution.

dead-end:
@willstott101 I am not sure that it is correct, that the repo contains an object as a loose object and in the packed file. Do you have some example code, that ends up in this situation?

willstott101:
According to Chat GPT :laughing: and these notes on `git gc --prune` sometimes loose objects are retained for 2 weeks by default when garbage collecting: https://git-scm.com/docs/git-gc#Documentation/git-gc.txt---pruneltdategt so even without network-related packs in upload-pack I believe it's valid to have objects in both places.

willstott101:
That being said --prune might only apply to orphaned loose objects:
```
.../test2.git # find objects/ | wc -l
1031
.../test2.git # git -C . gc --no-prune
Enumerating objects: 785, done.
Counting objects: 100% (785/785), done.
Delta compression using up to 2 threads
Compressing objects: 100% (431/431), done.
Writing objects: 100% (785/785), done.
Building bitmaps: 100% (12/12), done.
Total 785 (delta 458), reused 589 (delta 340), pack-reused 0
.../test2.git # find objects/ | wc -l
8
``

jcubic:
I know that the PR will fix this particular issue but I don't think that this is the root cause of the issue. I don't want to add a hack that doesn't solve the real issue.

[...]

PR #1790, opened 2023-07-10
fix: don't consider identical oids as ambiguous in expandOid (#1789)

willstott101:
As far as expandOid is concerned duplicate objects isn't actually ambiguous... so this seems like a sane fix.

~However IDK if the fact I have an object loose as well as in a pack file is something I should be concerned about.~

The existence of this command implies that duplicate objects are a valid (if sub-optimal) state for a git repo's blobs: https://git-scm.com/docs/git-pack-redundant

    You are better off using git gc to do so, which will put objects into a new pack, removing duplicates.


jcubic:
I don't think this is the right solution you just ignore that you have invalid data and return the first OID from duplicated ones.

willstott101:
I disagree this is invalid data. I believe it's perfectly possible to end up with loose objects that are also in packs. `writeObject` for instance just writes a loose object without holding a lock or anything afaict. This could happen alongside a git gc, or during a git push from another process etc... I don't see a way in the isomorphic api to atomically do write-if-not-exists, I am currently using:
`if (!expandOid) writeObject`

I'm also not convinced it'd be invalid for a duplicate to exist temporarily during a `git gc` from another process, for instance.

jcubic:
Sorry but if you can't provide a use case when this is broken, this is probably a useless change that only fixes some weird state you have in your app.

jcubic:
The release job failed. I'm not sure why not all tests passed. I've restarted the job, I hope it will pass and a release will be created.

PR #1799, opened 2023-08-02
fix: enable DecompressionStream detection

DenSpirit:
## I'm fixing a bug or typo


jcubic:
Can you explain what this fixes? It's not clear what is the purpose if this change.

Can you write a unit test that fails before the fix and works after the fix?

DenSpirit:
Hi. Sorry for taking so long to reply.
DecompressionStream support is supposed to be detected by isogit the first time it's requested. isogit checks for that by examining global libraries.
A value of `false` in `supportsDecompressionStream` implies that DecomressionStream detection was run and falsy, while `null` would imply that the detection was not run yet.

Because it's set to false in code, the detection never runs and isogit treats DecompressionStream as nonexisting (and defaults to pako) at all times.

If you look at a neighbouring file deflate.js, you'll see it has a line `let supportsCompressionStream = null`, which makes CompressionStream available for discovery. This merge request does the same for DecompressionStream.

DenSpirit:
I have added the unit tests, unfortunately they can't follow the regular style due to usage of local variables.
I am getting errors on certain browser versions, does it mean that the DecompressionStream support check is incomplete and more checks have to be done? I have already added a condition that was added to CompressionStream but not to DecompressionStream, probably that is not enough.

If possible, please assist :)

jcubic:
There are two issues:

1. recent changes in #1864 the stream needs to be closed
2. The tests are failing because of Android (probably some recent change with Chrome browser) I don't know what to do yet

jcubic:
The test got fixed, Please rebase and close the stream as on CompressionStream.

DenSpirit:
Made the necessary changes but it appears that the tests are still failing. I'm not sure the reason is my changes still, the failures seem unrelated.

jcubic:
It may be related to changes in Zlib library now that decompressionStream is actually used it produces differnet output. I'm not sure how since decompression should unpack old and new data in same way.

PR #1846, opened 2023-12-07
fix: add null check before comparing buffer length to cursor when looking for end of stream

mojavelinux:
If the server resets the connection (ECONNRESET), it's possible that the stream reader will be started but the buffer will be null. When this happens, the real error is masked by the following:

```
error TypeError: Cannot read properties of undefined (reading 'length')
    at StreamReader.eof
```

The solution is to verify buffer is not null before reading it. This is more than just a defensive check. There is a legitimate situation where the stream will be marked as ended but the buffer will be null.

### DIFF
--- a/src/utils/StreamReader.js
+++ b/src/utils/StreamReader.js
@@ -13,7 +13,9 @@ export class StreamReader {
   }
 
   eof() {
-    return this._ended && this.cursor === this.buffer.length
+    return (
+      this._ended && this.buffer != null && this.cursor === this.buffer.length
+    )
   }
 
   tell() {
--- a/src/utils/StreamReader.js
+++ b/src/utils/StreamReader.js
@@ -13,9 +13,7 @@ export class StreamReader {
   }
 
   eof() {
-    return (
-      this._ended && this.buffer != null && this.cursor === this.buffer.length
-    )
+    return this._ended && this.cursor === this.buffer.length
   }
 
   tell() {
@@ -23,8 +21,8 @@ export class StreamReader {
   }
 
   async byte() {
-    if (this.eof()) return
     if (!this.started) await this._init()
+    if (this.eof()) return
     if (this.cursor === this.buffer.length) {
       await this._loadnext()
       if (this._ended) return
@@ -34,8 +32,8 @@ export class StreamReader {
   }
 
   async chunk() {
-    if (this.eof()) return
     if (!this.started) await this._init()
+    if (this.eof()) return
     if (this.cursor === this.buffer.length) {
       await this._loadnext()
       if (this._ended) return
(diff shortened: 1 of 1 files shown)


mojavelinux:
There may be a better way to handle this. The eof check is happening too soon. Each method should first check if the stream needs to be started. Then it should check if the stream is at eof. This ensures that the buffer is always initialized and it avoids unnecessary _loadnext calls when the the stream started out ended.

I've added a commit with this improved solution. If we agree this is the right code, I'll squash it.

jcubic:
Is there a unit test that can be added that will test this? Something that will fail before the fix.

mojavelinux:
Despite the comment on the StreamReader class that it's extremely well tested, I found 0 tests for it (which is probably why this bug exists). I must admit that I don't entirely understand the call order when the reader is used. And while it could be unit tested, that would be artificial given it would only tell us that it works under that condition. The real test would be to emulate the connection being reset by the server.

I suppose the integration test would be to reproduce the situation that triggers the error, which is to configure the server with a very short timeout, then pause the connection for longer than that timeout, then try to resume it. While it's certainly possible to write such a test, it would take me more time than I have right now to figure out how to create it.

Btw, the reason this is now coming up is because GitHub has drastically reduce its connection timeout, so if a request is made, then paused for Node.js to service a separate concurrent operation (such as in Promise.all), GitHub will reset that connection.

When I have time, what I could do is create a script that reproduces the problem so it is clear the scenario we are trying to test.

jcubic:
No, not needed. If this is the problem with the server that is throwing an error, then there is no way to test it right now. We would need to add a completely new type of integration tests where we have our own server that we can control.

I think that you can safely marge the PR.

mojavelinux:
I'm doing some more low-level testing to both confirm the behavior and to explore whether a test is possible.

What I've discovered is that this only happens with POST requests (which isomorphic-git sends when negotiating various things such as references). And the limit seems to be 10 seconds (though I can't say that for absolute certain). If Node.js switches away from reading the response stream for more than this limit, GitHub will reset the connection and isomorphic-git will fail. While that failure is unavoidable, what we're trying to do here is make sure isomorphic-git doesn't report an undefined error but rather the connection reset error.

mojavelinux:
This fix on its own is not sufficient as it suppresses and underlying error (for example, an ECONNRESET error response). I'm working on a more comprehensive fix that allows such an error to bubble up. I may open a new PR depending on the nature of the change. Stay tuned.

mojavelinux:
> We would need to add a completely new type of integration tests where we have our own server that we can control.

FYI, this already exists. It's how the fetch and clone operations are tested.

mojavelinux:
I have tracked down the root causes and figured out how to both fix and test them. At this point, I'm going to create an issue to describe the problem and what we need to do in isomorphic-git to address it. I'll then submit a new PR. At that point, I'll probably close this one, but let me take one step at a time.

mojavelinux:
Superseded by the fix for #1847.

mojavelinux:
The change ends up being much simpler. If the stream is being marked as ended, and the value is falsy, the value should be set to Buffer.alloc(0). It's never going to be read after that point and it fixes the call to eof.

PR #1848, opened 2023-12-23
fix: rethrow error if error is encountered when reading git-upload-pack response

### LINKED ISSUE #1847: Properly propagate error encountered when reading response from git-upload-pack POST request
When performing a fetch operation on a repository (which occurs in any of following commands: clone, fetch, pull), isomorphic-git (as a compliant git client) will send a POST request to the git-upload-pack service to retrieve updates in the form of packetlines (new references) and the [packfile](https://git-scm.com/book/en/v2/Git-Internals-Packfiles) prepared by the server. A packfile is a payload that contains updates to the repository index (or the whole repository in the case of a clone). This response also includes progress summaries.

If an error occurs while reading the git-upload-pack response, instead of throwing that error, isomorphic-git swallows it, logs it to STDOUT, and presses on. In doing so, it either runs into an obtuse logic error (such as `Cannot read properties of undefined (reading 'length')`) or it writes an incomplete packfile, which manifests as an error later when reading objects in the repository.

This issue was originally reported as #1846. However, that was just the tip of the iceburg. isomorphic-git was trying to process an incomplete response (though it should still handle that situation gracefully).

The first question is, when does this happen? Well, a network connection is not a guaranteed resource. It can fail at any time for any reason. isomorphic-git makes the assumption that once it receives the 200 status code, it will be able to read the response to completion. But this is simply not a valid assumption.

In fact, we have received numerous reports of the `Cannot read properties of undefined (reading 'length')` over the years that are likely an indicator that the underlying error is being suppressed and ignored. The real reason is probably logged to STDOUT somewhere (if available), but that's very untidy for consumers of isomorphic-git since the error cannot be handled or redirected to a log.

But there's a more insidious problem afoot. GitHub recently changed its server settings so that any part of a response is not read for at least 5 seconds (or less if the client has opened multiple connections), it will reset the connection abruptly. This pause can happen if the isomorphic-git consumer runs concurrent fetch operations and at least of the responses is large. The pause is Node.js context switching between concurrent operations. If GitHub resets the connection when paused, Node.js will encounter an ECONNRESET error when resuming the stream. When this happens with the native git client, users will see the following error:

    fatal: unable to access 'http://localhost:8888/test-clone.git': transfer closed with outstanding read data remaining

Whether it's GitHub resetting the connection, or some other network problem, such as a client timeout, isomorphic-git has to stop processing the response when an error occurs and rethrow it so it can be handled.

I've found numerous places where isomorphic-git will either hang or throw an obtuse logic error if it encounters an error when reading some part of the response stream of a fetch operation:

* If the packfile is truncated (the lines after the PACK record header), StreamReader will try to access the `length` property of an undefined buffer, which results in `Cannot read properties of undefined (reading 'length')` (this was originally reported as #1846). (StreamReader#eof)
* If an error occurs before the PACK record header, isomorphic-git will log it to STDOUT, but proceed without a packfile, which may lead to the previous error or an error when reading the repository. (GitPackIndex.fromPack)
* If an error occurs before any lines are read from the response, isomorphic-git will hang because it does not resolve the Promise (parseUploadPackResponse)
* If an error record is encountered in the response itself, isomorphic-git will hang because it does not end the open streams (GitSideBand.demux)

When an error occurs when reading the stream, isomorphic-git should save that error to the `error` property on the stream instead of writing it to STDOUT (console.log).* (In GitSideBand.demux, it will also record the error on the packfile FIFO). The fetch command will then need to check for this property. It will need to do so in several places since isomorphic-git reads the response incrementally and may succeed with several steps (such as reading the packetlines) before hitting it.

I think we can address all of these problems in a single PR with multiple commits since they are all interconnected. I will submit that PR. (I have done rather extensive research into this problem and I'm confident that I know how to properly fix it).

(*) It's tempting to want to add an `error` listener on the response stream. Unfortunately, the http plugin in isomorphic-git does not require the value of the body object to honor the EventEmitter interface and therefore we cannot rely on it.

mojavelinux:
resolves #1847

* rethrow error if error is encountered when reading git-upload-pack response
* don't fail if packfile passed to GitPackIndex.fromPack is truncated
* avoid unnecessary copy of buffer when writing line to packetlines FIFO
* properly close FIFO objects after encountering error line in sideband response
* capture error in GitPktLine#read instead of logging to stdout

### DIFF
--- a/__tests__/test-GitPackIndex.js
+++ b/__tests__/test-GitPackIndex.js
@@ -70,6 +70,17 @@ describe('GitPackIndex', () => {
       20855
     )
   })
+  it('from .pack when pack is truncated', async () => {
+    const { fs, gitdir } = await makeFixture('test-GitPackIndex')
+    const pack = await fs.read(
+      path.join(
+        gitdir,
+        'objects/pack/pack-1a1e70d2f116e8cb0cb42d26019e5c7d0eb01888.pack'
+      )
+    )
+    const p = await GitPackIndex.fromPack({ pack: pack.slice(0, 12) })
+    expect(p.offsets.size).toBe(0)
+  })
   it('to .idx file from .pack', async () => {
     const { fs, gitdir } = await makeFixture('test-GitPackIndex')
     const idx = await fs.read(
--- a/__tests__/test-GitPktLine.js
+++ b/__tests__/test-GitPktLine.js
@@ -73,6 +73,37 @@ describe('GitPktLine', () => {
     )
   })
 
+  it('read stream - with error', async () => {
+    const hookStream = (subject, fn) => {
+      const unhook = function(write) {
+        this.write = write
+      }.bind(subject, subject.write)
+      subject.write = fn
+      return unhook
+    }
+    let unhook
+    try {
+      const output = []
+      if (!process.browser) {
+        const onLog = chunk => output.push(chunk) && undefined
+        unhook = hookStream(process.stdout, onLog)
+      }
+      let err
+      const stream = {
+        next() {
+          throw (err = new Error('something went wrong'))
+        },
+      }
+      const read = GitPktLine.streamReader(stream)
+      const line = await read()
+      expect(output.length === 0).toBe(true)
+      expect(line).toBe(true)
+      expect(stream.error === err).toBe(true)
+    } finally {
+      if (unhook) unhook()
+    }
+  })
+
   it('encode empty string', async () => {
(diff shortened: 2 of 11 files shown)


jcubic:
Looks good.

mojavelinux:
:tada:

I'm ready to merge when you are. I've manually tested this change in my application and it works just as it's intended.

jcubic:
Ok then merging.

mojavelinux:
:tada:

PR #1853, opened 2024-01-06
fix: `normalizePath` performance

### LINKED ISSUE #1852: Improve `normalizePath` performance
From my benchmark I see the code is spending most of its time in the [`normalizePath`](https://github.com/isomorphic-git/isomorphic-git/blob/ad9bc11e2df311da540330d2243cb3ba9dd5d47a/src/utils/normalizePath.js#L1) function.
When I looked at it I found it is doing multiple Regex in sequence.
This can be optimized with multiple ways

I had a ~800% performance improvement on very large repository using this `normalizePath` implementation:
```javascript
function normalizePath(path) {
  path = path.split('/./').join('/').replace(/\/{2,}/g, '/');
  
  if (path === '/.') return '/';
  if (path === './') return '.';
  if (path.startsWith('./')) path = path.slice(2);
  if (path.endsWith('/.')) path = path.slice(0, -2);
  if (path.length > 1 && path.endsWith('/')) path = path.slice(0, -1)

  return path || '.';
}
```

We could also compare this implementation with a compiled Regex implementation and matching

I think it could be improved further with a very good knowledge of the library.
This is why I do not propose a Pull Request, I think it should be discussed before decided.
What do you think guys ?

scolladon:
I propose a performance optimization for the `normalizePath` function.

I use `isomorphic-git` to refactor a sfdx plugin I maintained and I was very happy with it as it was running 3 times faster for small to medium repository.
Then I tried with a very big one and I found the performances were dropping.
So I looked at what happened to see where I could tweak my implementation (maybe I can have a smarter `iterate` or `walk` method tailored made).
This is how I discovered that 70% of the time were spent inside the `normalizePath` function.

This PR tries to have a very simple approach and to avoid using regex for this code massively used by the library.
There may be a smarter way to deal with it (do multiple things at the same time, skip things), please advise !

On my specific use cases, this `normalizePath` implementation made my refactoring 800% faster than without

closes #1852

### DIFF
--- a/src/utils/normalizePath.js
+++ b/src/utils/normalizePath.js
@@ -1,4 +1,14 @@
 export function normalizePath(path) {
+  let normalizedPath = normalizePathCache.get(path)
+  if (!normalizedPath) {
+    normalizedPath = normalizePathInternal(path)
+    normalizePathCache.set(path, normalizedPath)
+  }
+  return normalizedPath
+}
+
+const normalizePathCache = new Map()
+function normalizePathInternal(path) {
   path = path
     .split('/./')
     .join('/') // Replace '/./' with '/'
--- a/src/utils/normalizePath.js
+++ b/src/utils/normalizePath.js
@@ -1,11 +1,14 @@
 export function normalizePath(path) {
-  return path
-    .replace(/\/\.\//g, '/') // Replace '/./' with '/'
+  path = path
+    .split('/./')
+    .join('/') // Replace '/./' with '/'
     .replace(/\/{2,}/g, '/') // Replace consecutive '/'
-    .replace(/^\/\.$/, '/') // if path === '/.' return '/'
-    .replace(/^\.\/$/, '.') // if path === './' return '.'
-    .replace(/^\.\//, '') // Remove leading './'
-    .replace(/\/\.$/, '') // Remove trailing '/.'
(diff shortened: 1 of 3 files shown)


jcubic (commented):
Thanks for the PR, I will check the old implementation in more detail maybe tomorrow to see if they are the same. All tests pass so it's promising.

jcubic:
Thanks for the reminder, I completely forget about this.

src/utils/normalizePath.js
```diff
@@ -1,11 +1,27 @@
+const memo = new Map()
```
jcubic, line 1: What is the use of this memo variable?

jcubic, line 1: Is it really important? This will grow and use memory. Maybe using WeakMap here will be a better option. What do you think?

scolladon: I'm not sure WeakMap would be appropriate as it needs object, but I see your point.

If memory consideration is important then I think the improvement without the memoization is enough.

If we can use memory then just memoizing this method improve performances further (+10%). The cost of memoizing this method memory wize is not that huge because the map contains only strings.

Your call, I don't mind removing it :)

jcubic: Let's keep it and if it will consume too much memory we will remove it later.

Issue #1851, opened 2024-01-05
isPromiseFs fails on Deno


mdekstrand:
When attempting to run in Deno (with `node:fs` as the fs), attempting a git operation fails with the following error:

```
error: Uncaught (in promise) NotFound: No such file or directory (os error 2): readfile ''
    at async Object.readFile (ext:deno_fs/30_fs.js:720:18)
```

When I can get a full stack trace, this is appearing through `isPromiseLike()`. Testing at the Deno REPL indicates that error when `readFile()` is called without arguments is escaping the `catch` clause. I am not sure what would be causing this on the Deno end, but REPL testing indicate that explicitly passing an empty string argument to `readFile()` would probably correct the problem.

jcubic:
Do you want to contribute and create a fix?

As for the catch in Deno, maybe this should reported as a bug for Deno. Errors should not escape catch unless the error is retrown.

mdekstrand:
I would definitely be open to contributing a fix. I've also reported a bug against Deno (denoland/deno#21795), and will see what they say there first. It definitely seems like a bug, even if it's really weird to call `readFile()` without an argument.

mdekstrand:
@jcubic Thinking about this, is there a reason it needs to check that `fs.promises` is actually promise-like? I see why it needs to check if you just pass `fs`, but could that check be dropped without causing other problems?

Alternatively, would it work to provide a way to say “this one is really a promise-based api, trust me”?

jcubic:
The problem is that the library expects to detect if it's normal FS or promisified FS. If it's normal FS it promisify all the functions from FS. So yes it needs to check if the result is fs.promise or not. But maybe it can check different method that don't throw an error.

I'm wondering if `fs.stats` will work on a root directory. It needs to work for both browser and Node (and in your case Deno). It needs to be a method that is included in [lightningFS](https://github.com/isomorphic-git/lightning-fs).

jcubic:
You can also report this bug to Deno. Exceptions should not bypass `try..catch`.

mdekstrand:
It's being discussed over at Deno, and this comment gets to what's happening here: https://github.com/denoland/deno/issues/21795#issuecomment-1903901348.

[...]

PR #1857, opened 2024-01-22
fix: isPromiseFs out-of-spec usage of readFile

mdekstrand:
This works around the odd Deno behavior described in #1851 by explicitly passing `undefined` to `readFile`, so the `promisify` wrapper works correctly.

This is arguably a Deno bug (denoland/deno#21795), but it's also an emergent behavior of isomorphic-git calling `readFile` out-of-spec + promisfy's inability to identify that because it is unaware of intended method signatures. This workaround gets `isPromsseFs` working on Deno, and that + manually adding `Buffer` to `globalThis` is enough to get isomorphic-git working on Deno for my use case.

Full Deno support (including running the tests) would be a much larger effort.

## I'm fixing a bug or typo

### DIFF
--- a/src/models/FileSystem.js
+++ b/src/models/FileSystem.js
@@ -10,7 +10,7 @@ function isPromiseFs(fs) {
     try {
       // If readFile returns a promise then we can probably assume the other
       // commands do as well
-      return targetFs.readFile().catch(e => e)
+      return targetFs.readFile(undefined).catch(e => e)
     } catch (e) {
       return e
     }


mdekstrand:
Test failures are a couple of Chrome pack tests that I don't think this code touched.

jcubic:
It fails for Android, the the code touched is always executed so the error came from that code.
Can you try changing it to an empty string?

mdekstrand:
Changing to `''` those two tests still fail, and now a 3rd does as well. I will revert to the undefined version when I get home.

The failing assertion in both versions is that the pack name does not match the expected pack name.

jcubic:
The last PR was this one [#1853](https://github.com/isomorphic-git/isomorphic-git/pull/1853) and the test was passing.

I will try to create a trivial PR to test if the tests pass.

jcubic:
The tests are broken, I have no idea why. Here is a README change #1858 that is also failing.

I will merge this PR when I fix the tests, never happen before. I hope I will be able to figure out why the tests are failing.

jcubic:
The tests got fixed, please rebase.

Issue #1859, opened 2024-01-22
Tests are failing on Android


jcubic:
The PR that add change to README is failing in Android: #1858

jcubic:
I think that we need to figure out which service is failing Azure, SauseLabds, or BrowserStack and ask support about this.

inverted-capital:
this is fixed in PR #1864

jcubic:
The real problem is that the zlib compress library was updated in Chrome 120 on Android. And the test compare binary blob of pack file. So the tests need to be update to decompress the data before comparing. This is what Chrome team suggested, to not compare binary compression blobs.

I attempted to fix the tests, but I'm not sure how pack files are created, locally all tests pass but on CI there are even more failing tests.

I run the test with:

```
npx nps test.setup && npx nps test.jest
```

inverted-capital:
Pardon my repetition, but I think it is a bad idea to abandon binary compatibility checks just because a single platform has recently failed a compatibility check that has stood for nearly 5 years, and still stands on all other platforms.

jcubic:
We can still test binary compatibility if the compression algorithm has a stable API. Decompressing old and new stuff gives the same results. We can have a workaround but call deflate on the packfile and not worry about comparing binary blobs.

inverted-capital:
But, surely that would not test the case where a new packfile was created on Android, and then opened on Ubuntu FireFox, for example ?  If the file format on android has changed, we cannot be sure it will decompress on all the other platforms, unless we are sure it is binary equal, right ?

jcubic:
I got a reply from the Zlib author he said that they don't know of any recent change in the output. But he also said that it's not guaranteed that the binary data will always be the same. The only guarantee is that they `decompress(compress(x)) == x` and that decompress will always be backward compatible. So we can keep the binary blob as a fixture and decompress both the new pack file and fixture. and compare data inside. I only need to check the read data from the pack file using isomorphic-git.

inverted-capital:
That is nice of him to respond so quickly.

zlib (or any other format) cannot be forwards compatible unless it is binary equal.

If `compressNew()` is running on android with this changed format, and `decompressOld() is running on another platform that is older than the new Android format, then the following is NOT guaranteed:

`decompressOld( compressNew( x ) ) == x`

I will leave this issue be - hopefully no further issues come up related to it.  It is really weird that it changed on Android at all.

jcubic:
[...]

PR #1862, opened 2024-01-26
fix: handle no-refs response from git-upload-pack for git > 2.41.0

### LINKED ISSUE #1860: Cloning empty repository fails if git server uses git >= 2.41.0
Although [it's been documented for 15 years](https://github.com/git/git/blame/master/Documentation/gitprotocol-pack.txt#L212-L222), git was recently changed to align with the gitprotocol-pack specification in the docs (used by git-upload-pack). In brief, if the remote repository has no refs, git < 2.41.0 sends back the following null response.

```
001e# service=git-upload-pack
0000
```

However, the specification states that it should do the following:

```
no-refs = PKT-LINE(zero-id SP "capabilities^{}" NUL capability-list)
capability-list = capability *(SP capability)
zero-id = 40*"0"
```

Starting with git 2.41.0 (https://github.com/git/git/commit/933e3a4ee205353d8f093d5dfcd226fa432c4e58), git now sends a zero id with a symbolic reference followed by the list of capabilities (as the aforementioned specification states).

```
001e# service=git-upload-pack
00000000000000000000000000000000000000000000 capabilities^{}\0multi_ack thin-pack side-band side-band-64k ofs-delta shallow deepen-since deepen-not deepen-relative no-progress include-tag multi_ack_detailed no-done object-format=sha1 agent=git/2.43.0
```

When isomorphic-git parses this response, it incorrectly assumes that the first ref is named "capabilities^{}" with a zero sha (see https://github.com/isomorphic-git/isomorphic-git/blob/main/src/wire/parseRefsAdResponse.js#L42). It then proceeds to try to find this refname, which is fails to do. It should have already returned at this point with an empty set of refs (see https://github.com/isomorphic-git/isomorphic-git/blob/main/src/commands/fetch.js#L126-L127).

This issue is not yet reproducible on a GitHub repository since GitHub is not yet using git >= 2.41.0 (though we don't know which version since it is masked). However, it can be reproduced by cloning an empty git repository on GitLab. Here's an example:

```js
const git = require('isomorphic-git')
const http = require('isomorphic-git/http/node')
const fs = require('node:fs')

;(async () => {
  dir = '/tmp/empty.git'
  fs.rmSync(dir, { recursive: true, force: true })
  await git.clone({
    fs,
    http,
    url: 'https://gitlab.com/mojavelinux/empty.git',
    corsProxy: false,
    depth: 1,
    dir,
    noCheckout: true,
    noTags: true,
  })
})()
```

The fix is to check for the special capabilities line when reading the git-upload-pack response for an empty repository. I'll submit a fix with a test case.

mojavelinux:
closes #1860

### DIFF
--- a/__tests__/__fixtures__/test-empty.git/description
+++ b/__tests__/__fixtures__/test-empty.git/description
@@ -1,0 +1,1 @@
+a git repository with no refs; the objects and refs folders are required to make it a valid repository
--- a/src/wire/parseRefsAdResponse.js
+++ b/src/wire/parseRefsAdResponse.js
@@ -41,14 +41,17 @@ export async function parseRefsAdResponse(stream, { service }) {
 
   const [firstRef, capabilitiesLine] = splitAndAssert(lineTwo, '\x00', '\\x00')
   capabilitiesLine.split(' ').map(x => capabilities.add(x))
-  const [ref, name] = splitAndAssert(firstRef, ' ', ' ')
-  refs.set(name, ref)
-  while (true) {
-    const line = await read()
-    if (line === true) break
-    if (line !== null) {
-      const [ref, name] = splitAndAssert(line.toString('utf8'), ' ', ' ')
-      refs.set(name, ref)
+  // see no-refs in https://git-scm.com/docs/pack-protocol#_reference_discovery (since git 2.41.0)
+  if (firstRef !== '0000000000000000000000000000000000000000 capabilities^{}') {
+    const [ref, name] = splitAndAssert(firstRef, ' ', ' ')
+    refs.set(name, ref)
+    while (true) {
+      const line = await read()
+      if (line === true) break
+      if (line !== null) {
+        const [ref, name] = splitAndAssert(line.toString('utf8'), ' ', ' ')
+        refs.set(name, ref)
+      }
     }
   }
   // Symrefs are thrown into the "capabilities" unfortunately.


jcubic:
This can't be merged until Unit Tests for Android are fixed. See https://github.com/isomorphic-git/isomorphic-git/issues/1859

jcubic:
I think that we need to figure out which service is failing Azure, SauseLabds, or BrowserStack and ask support about this.

jcubic:
Can you rebase (or `commit --amend --no-edit`) and push force so the tests run again?

mojavelinux:
Done.

jcubic:
Ok, I'm merging the unit tests for this will be added in https://github.com/isomorphic-git/isomorphic-git/pull/1895 (someone provided the same fix, and he added tests).

jcubic:
Next time you should create a branch in this repo, the author of that other PR would be able to create add it to your branch (I think it's called stacked PR).

jcubic:
The release failed. Not sure what to do. @seanpoulter can you take a look? The Job take more than 30 minutes and get stopped by Azure. Here is a link:

https://dev.azure.com/isomorphic-git/isomorphic-git/_build/results?buildId=3376&view=results

seanpoulter:
Is that the first attempt? I'd retry if you can. If you enable timestamps in the logs it seems like our test code was running in the usual amount of time but it stalled after timing out trying to connect to `104.19.163.75` (BrowserStack).

jcubic:
I clicked rerun failed jobs, I hope it will run everything.

seanpoulter:
Yay, that looks like it worked and updated the website too.

I'm going on intuition here but I'd assume it's the `karma-browserstack-launcher` that interacts with BrowserStack. We're using v1.3.0 but v1.6.0 is available, and the repo was archived about a year ago. It's another good signal to rework the test framework.

mojavelinux:
> Next time you should create a branch in this repo

I dislike that pattern as it pollutes the upstream repo with branches. The typical workflow is to submit PRs from branches. It should be possible to push to that branch if you are a maintainer of this repository. Otherwise, you can submit a PR to that branch and when it is merged, it will be incorporated into the upstream.

For me, it's no trouble to be asked to rebase and force push. That is all part of the workflow of a PR in my mind.

jcubic:
I was only thinking to make PR to PR easier.

PR #1865, opened 2024-02-18
fix: close the test compression stream

inverted-capital:
Turns out that resources are still reported as leaking in Deno if the Compression stream is not closed also.

### DIFF
--- a/src/utils/deflate.js
+++ b/src/utils/deflate.js
@@ -22,12 +22,12 @@ async function browserDeflate(buffer) {
 function testCompressionStream() {
   try {
     const cs = new CompressionStream('deflate')
+    cs.writable.close()
     // Test if `Blob.stream` is present. React Native does not have the `stream` method
     const stream = new Blob([]).stream()
     stream.cancel()
-    return !!cs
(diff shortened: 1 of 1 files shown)


src/utils/deflate.js
```diff
@@ -22,12 +22,12 @@ async function browserDeflate(buffer) {
...
     const cs = new CompressionStream('deflate')
+    cs.writable.close()
     // Test if `Blob.stream` is present. React Native does not have the `stream` method
     const stream = new Blob([]).stream()
     stream.cancel()
-    return !!cs
+    return true
   } catch (_) {
-    // no bother
+    return false
```
jcubic, line 31: return false is already handled in next lines

inverted-capital: I removed return false from outside the try catch block since the only way false is returned is if an error is thrown.

An empty catch block is frowned upon by linters, which is presumably why the comment was put there, to silence linter errors.

jcubic: Ah, didn't notice that the return was removed.

PR #1877, opened 2024-03-07
codespell CI: Remove a single github workflow , move into azure-pipelines.yml

### LINKED ISSUE #1875: Codespell workflow needs approval
Issue with Code spell from #1843. Each time when someone creates a PR, maintainer need to approve running a workflow. Which is silly when unit tests run just fine automatically.

yarikoptic:
currently running codespell requires approval since I guess there are some secrets set for the repo.

I made it last in the azure-pipelines.yml so there is no interference of any kind on prior "non-pythonic" steps

Related
- #1496 

Closes #1875 

## I'm fixing a bug or typo


seanpoulter:
Is there another solution we should consider? The goal of #1496 is to only use GitHub Actions.

yarikoptic:
> Is there another solution we should consider? The goal of #1496 is to only use GitHub Actions.

but it is not the case ATM and you are using azure -- so I decided to harmonize into it.

I have not found relevant information -- seems like you have some github secrets for actions and that is why an action from a personal fork seems to require approval even if not using any of the secrets.  @jwodder -- do you know more on this?  codespell workflow here requires approval although not using any secrets -- any way to overcome without some complicated additional bot/action to auto-approve it?

jwodder:
@yarikoptic [This is just how GitHub works](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks):

> [W]orkflows on pull requests to public repositories from some outside contributors will not run automatically, and might need to be approved first. By default, all first-time contributors require approval to run workflows.

jcubic:
Ah, ok. Didn't know about it. So we can ignore this. Sorry for the trouble. I guess we will need to accept all actions when we migrate to GH workflows. I thought it was some kind of mistake that could be fixed.

seanpoulter:
We could pivot a bit and document the behaviour so contributors know what to expect. What do you think?

jcubic:
Good idea, let's do this after migration to GitHub actions.

Issue #1881, opened 2024-03-14
Race condition makes it possible for a ref to be resolved to an empty string while it's in the middle of being updated


alexvy86:
We're encountering this problem while using isomorphic-git in a Node 18 environment.

I don't have a good minimal reproduction yet but can reproduce it more or less consistently with a big suite of tests of https://github.com/microsoft/FluidFramework (I'm part of the team working on it).

Specifically, a call to [`GitRefManager.writeRef()`](https://github.com/isomorphic-git/isomorphic-git/blob/79a38e017653a4dfe3b93e390bf3ffa73db42e82/src/managers/GitRefManager.js#L137) that happens while a call to [`GitRefManager.resolve()`](https://github.com/isomorphic-git/isomorphic-git/blob/79a38e017653a4dfe3b93e390bf3ffa73db42e82/src/managers/GitRefManager.js#L179) (for the same ref) is underway sometimes causes [this read](https://github.com/isomorphic-git/isomorphic-git/blob/79a38e017653a4dfe3b93e390bf3ffa73db42e82/src/managers/GitRefManager.js#L203) in the `resolve()` call to return an empty string if it happens around the same time as [this write](https://github.com/isomorphic-git/isomorphic-git/blob/79a38e017653a4dfe3b93e390bf3ffa73db42e82/src/managers/GitRefManager.js#L142). A second `resolve()` call immediately after returns the correct value written by the update, which supports the theory that it's a concurrency issue with the read and write.

I couldn't find docs that suggest this is a known possible issue or any guidance on how to handle this scenario, so I suspect it might be a bug.

Assuming it is, I can think of two solutions off the top of my head: stop using the Promise-based APIs in the provided `fs` module in favor of sync ones (which feels pretty drastic and probably a noticeable performance hit), or add locking around write/update operations (also a performance hit, maybe not as bad). Neither seems ideal, but the correctness problem seems like it needs to be addressed somehow...

jcubic:
Do you want to contribute with a fix? I think that locking when writing/resolving refs is a good idea.

alexvy86:
@jcubic [first draft is out](https://github.com/isomorphic-git/isomorphic-git/pull/1882). I think the general approach is there, I'd still like a bit of time to make sure I'm thinking everything through and determine if any cases don't need locking (like right now I skipped it for `listRefs` even though it uses the filesystem, since it only cares about listing the files, not looking at its contents).

alexvy86:
[...]

PR #1882, opened 2024-03-14
Fix race condition when reading/writing refs

alexvy86:
## Description

Adds locking around filesystem operations when reading/writing refs to avoid race conditions.

Fixes https://github.com/isomorphic-git/isomorphic-git/issues/1881.

## I'm fixing a bug or typo

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -669,6 +669,15 @@
       "contributions": [
         "doc"
       ]
+    },
+    {
+      "login": "alexvy86",
+      "name": "Alex Villarreal",
+      "avatar_url": "https://avatars.githubusercontent.com/u/716334?v=4",
+      "profile": "https://alex-v.blog/",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "commitConvention": "angular"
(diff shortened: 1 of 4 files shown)


jcubic:
It would be nice to have a unit test for this. I know you said it's hard to reproduce the issue, but maybe do something with threads and try to save and read at the same time.

If you will not be able to tests the locks, we will at least to make sure they don't interfere with normal workflow.

And about the Unit tests, they are failing because of broken credentials in Azure DevOps. We are in the process of migrating the CI/CD to GitHub actions, but maybe we will disable the failing tests for now, so we will be able to test.

jcubic:
Looks good.

So now we need to fix the tests. Probably disable failing ones temporarily. I will be not available during a weekend, will get back to this on Monday. The tests are important because they block contribution.

alexvy86:
I noticed other places were already using the `async-lock` package so just used that too for this fix, and managed to write a unit test _that is not deterministic_, but as far as I could tell, reproduced the problem pretty reliably before applying the fix.

I see there's already https://github.com/isomorphic-git/isomorphic-git/pull/1884 working on disabling the failing test. Not to be pushy, but isn't there someone with admin permissions on the repo that could merge contributions even with failed checks? As long as that is the only failing test, that seems safe to me.

jcubic:
I have admin rights and can merge even when test fail, but I don't want to. Also, I think that the release will not happen when the tests are failing. CI/CD will stop execution if there is an error. But even it this would work, it's not good practice to ignore the tests and merge anyway. Tests are created for a reason. To make sure the code works, not something you can just ignore for your convenience.

jcubic:
The failing test got merged, please rebase.

alexvy86:
> To make sure the code works, not something you can just ignore for your convenience.

Absolutely, and if the failure had been due to a code problem, that's definitely a reason not to merge. Since in this case it was clear that the failing test was not because of a problem in code but an issue with an external system, my take is that it would have been ok to do. But again, not trying to push it, just giving my opinion :).

I'll rebase now!

PR #1886, opened 2024-03-21
Update statusMatrix.js: added two more StatusMatrix example

tmpmachine:
## I'm editing the documentation
Affected page: https://isomorphic-git.org/docs/en/statusMatrix

Added two more StatusMatrix examples:
```
 *   ["j.txt", 1, 2, 0], // deleted, staged, with unstaged-modified changes (new file of the same name)
 *   ["k.txt", 1, 1, 0], // deleted, staged, with unstaged changes (new file of the same name)
```

This status is for when you delete a file, then create a new one of the same name (still in working directory).

I'm not sure what to edit on the "possible combinations" part where it marked with `??`. Maybe something like this?

```
0	 2	0	 ??       -> U   (Untracked)
1	 1	0	 D + ??   -> DU  (Deleted, Untracked)
1	 2	0	 D + ??   -> DM  (Deleted, Modified)
```

# Canonical Git Scenario

In both scenario, we have a file in HEAD, deleted it, add it to INDEX, then created a new file of the same name. The difference is that the file in scenario 1 has content that identical to HEAD, and scenario 2 has content that differs from HEAD.

1. StatusMatrix 1,1,0 (Deleted, Untracked)
```bash
touch README.md
git add README.md
git commit -m "added readme"
rm README.md
git add README.md
touch README.md
```

2. StatusMatrix 1, 2, 0 (Deleted, Modified)
```bash
touch README.md
git add README.md
git commit -m "added readme"
rm README.md
git add README.md
touch README.md
echo "Hello, world!" > README.md
```

### DIFF
--- a/src/api/statusMatrix.js
+++ b/src/api/statusMatrix.js
@@ -57,6 +57,8 @@ import { worthWalking } from '../utils/worthWalking.js'
  *   ["g.txt", 1, 2, 3], // modified, staged, with unstaged changes
  *   ["h.txt", 1, 0, 1], // deleted, unstaged
  *   ["i.txt", 1, 0, 0], // deleted, staged
+ *   ["j.txt", 1, 2, 0], // deleted, staged, with unstaged-modified changes (new file of the same name)
+ *   ["k.txt", 1, 1, 0], // deleted, staged, with unstaged changes (new file of the same name)
  * ]
  * ```
  *


tmpmachine:
@jcubic done, added canonical git scenario. I'll let others decide the label used in "possible combinations" table.

jcubic:
This is just documentation, so I think it's fine.

PR #1894, opened 2024-03-28
Set up for AMRC fork

amrc-benmorrow:
* Add a section to the README documenting this fork.
* Update `package.json` to use the new name and version.
* Remove steps we don't need from build and CI.


amrc-benmorrow:
Sorry, wrong repo. This was meant to go to a fork.

jcubic:
Hey @amrc-benmorrow if the fork is to "fix a critical bug", why don't you contribute with a fix to the repository. You will not need to maintain the fork.

What kind of critical bug needs fixing?

amrc-benmorrow:
Hi Jakub,

We are intending to contribute our fix back soon. Right now we are in the
middle of the release process for a large software system and we need a
risk-free way to get this fixed without waiting.

The fix is specifically to handle the situation where an empty repository
is cloned and the server returns capabilities. Somewhere between git
v2.40.1 and v2.43.1 git-http-backend has started returning capabilities for
an empty repo; this is a special case in the protocol which isomorphic-git
doesn't currently handle correctly. A routine update to our docker base
images has led to the git binaries being updated which has made this come
to light.

I'm glad to see the project is still actively maintained; it wasn't clear
to me from the Github page whether a PR would be accepted, or how long it
might take to make it into a release we could use. I will submit a PR once
I get a chance.

Ben

PR #1909, opened 2024-05-05
fix: handle non utf8 files with autocrlf correctly

### LINKED ISSUE #1908: Unchanged image/audio files being queued for commit with `git.add()`
isomorphic-git: 1.25.8
Node: 20.12.2

This behaviour seems to have changed since upgrading from 1.25.7.

I have a repository with a few modified files (all text files). I'm then running `git.add()`:
```typescript
await git.add({
    fs,
    dir: this.repoPath,
    filepath: '.'
});
```

My expectation is that this will add all modified files (only) for commit. However, it is adding not only those files but also some (perhaps all) image and audio files for commit also - even though these are unmodified.

Reverting to 1.25.7 returns to the expected behaviour.

jayree:
fix #1908


jcubic:
Can you add a unit test for this?

jayree:
@jcubic I tried, but I don't know how to create a new fixture with a binary file included.

jcubic:
You can try to include files that contain `\r` and different files that changed. When adding all of them, only files that changed should be in the staging area by statusMatrix. It can be a test for StatusMatrix.

jayree:
@jcubic the problem is not the test case, I'm unable to create a new working fixture repository. If I include the `index` the `workdir` is empty. If I remove the index all files have the status `*added`

jcubic:
I will check later today on how to create fixures.

jayree:
@jcubic I've added a test which builds the test case from scratch

jcubic:
Are those hooks even used? It looks like sample code created by git init. Can you remove them?

jayree:
@jcubic Is anything else missing?

jcubic:
I would add yet another test, to make sure that normal text files with \r that don't change also are not marked as modified. You can copy and paste the test and use a normal text file instead of a binary file.

jayree:
@jcubic done

PR #1925, opened 2024-06-10
fix: support autocrlf input

jayree:
Add the missing handling for `core.autocrlf=input`


jcubic:
According to [docs](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_core_autocrlf) `input` is different from `true`. Are you sure this s correct?

PR #1932, opened 2024-06-13
fix: Use native inflate if available

elegaanz:
Because `supportsDecompressionStream` was set to `false` by default, the test to see if native inflate was available or not was actually never run.


elegaanz:
I don't know if I should squash my two commits together or not, please tell me what you prefer.

jcubic:
thanks for the PR, it need to wait a bit until we fix the build pipeline.

jcubic:
No, you don't need to squash. I will do that when merging. I should probably remove that point from PR template.

jcubic:
I disabled broken GitLab test, can you rebase?

mojavelinux:
I agree that this seems to have been a coding error. The code was added but never used.

In order for this to actually work when running tests in Node.js, the following needs to be added to jest.config.cjs as well:

```
  globals: {
    Blob: global.Blob,
    CompressionStream: global.CompressionStream,
    DecompressionStream: global.DecompressionStream,
    Response: global.Response
  },
```

This will be related to the fix for #1933.

mojavelinux:
Hold on. I just did some quick performance testing and determined that using the DecompressionStream in Node.js performs terribly when compared to pako (and even worse when compared to the built-in zlib). This would be a very bad change to make now if it truly does have that consequence. On one large repository, switching to DecompressionStream caused the kernel to kill the Node.js process. It's possible the there is a leak in the isomorphic-git code as well.

Issue #1240, opened 2020-10-11
readObject with type = wrapped throws for a commit object


chmac:
Example reproduction:

```typescript
        const wrappedObject = await git.readObject({
          fs,
          oid: 'objectId of a commit',
          format: 'wrapped',
        });
```

When I try the above, I get the following exception:

```
TypeError: Cannot set property 'oid' of undefined
    at Object.readObject (node_modules/isomorphic-git/index.cjs:11827:16) {
  caller: 'git.readObject'
```

Digging into this a bit, I find that here `result` is undefined:

https://github.com/isomorphic-git/isomorphic-git/blob/f2e38053733f4c8ca4b7784d5c2a56addcf36bf4/src/api/readObject.js#L228-L235

It seems like `_readObject()` is returning `undefined`. Maybe there's a bug in this switch statement?

https://github.com/isomorphic-git/isomorphic-git/blob/071e8ec92f8f0e07a8a067082f0a0f316f3ba026/src/storage/readObject.js#L61-L70

Going into this switch statement, `format = 'wrapped'` while `result.format = 'content'`. So we hit the `case 'content':`, but then not the nested `if`, which means we hit `break`, and so we skip the `default` branch, but there's no closing `return` in the function, and so `result` is never returned, and we get back `undefined` instead.

I'm not sure what's going on here. Perhaps for a commit object, `wrapped` and `content` are synonymous because there is no "wrapped" version of a commit?

jcubic:
Hey, thanks for reporting a bug and great production and find the root cause. Maybe you can create a PR with a fix.

Sorry for the late reply, but I've become a maintainer recently and and I've started to handling issues and PR. The original author doesn't have much time to work on the project so I've started helping, but since I also don't have much time I don't know if I will be able to write code. So if you want to help and become contributor to this Open Source project, your help is welcome.

balajianbalagan:
Hey, can I work on this issue? I'm new to Open source contribution but good at JS and programming in general

jcubic:
Go ahead, if you can fix the issue that would be great. But I'm not sure what would be the right code to fix the issue. You will need to figure it out on own, which is the most difficult part of the issue.

dead-end:
@balajianbalagan I think the code changed since the issue was reported. I get the mentioned error:

```
TypeError: Cannot set properties of undefined (setting 'oid')
```
when I pass a wrong format:
```
        const wrappedObject = await git.readObject({
          fs,
[...]

PR #1935, opened 2024-06-22
Fix readObject when reading from packed objects with format `wrapped`

limond:
Relates to #1240 

**Full bug description**: https://github.com/isomorphic-git/isomorphic-git/issues/1240#issuecomment-2184005818
**TLDR**:

This PR fixes a bug where it was previously not possible to call `readObject` with format `wrapped` for packed objects. An error similar to this would throw instead:
```
TypeError: Cannot set property 'oid' of undefined
    at Object.readObject (node_modules/isomorphic-git/index.cjs:11827:16) {
  caller: 'git.readObject'
  ```

The [documentation](https://isomorphic-git.org/docs/en/readObject) states:
> format is usually, but not always, the format you requested. Packfiles do not store each object individually compressed so if you end up reading the object from a packfile it will be returned in format 'content' even if you requested 'deflated' or 'wrapped'.

This was not true for `wrapped` before.

This PR also adds unit tests for reading `deflated`, `wrapped` or `content` objects from a packfile (previously only `deflated` was present).

This PR introduces early returning for packed objects in `_readObject` and cleans up a few `if` statements that can be simplified after early returning.


jcubic:
I disabled broken GitLab test, can you rebase?

PR #1939, opened 2024-06-24
fix: transform response from deflate compression stream to work with Node.js 22.3.0

### LINKED ISSUE #1933: Reading commit with node 22.3.0
Using node with version 22.2.0 all works perfectly, but using version 22.3.0 (that is the latest) I got the following error when trying to read a commit:
```
TypeError {
  message: 'Cannot create property \'caller\' on string \'incorrect header check\'',
}

TypeError: Cannot create property 'caller' on string 'incorrect header check'
    at Object.readCommit (/home/matteo/src/slangroom/node_modules/.pnpm/isomorphic-git@1.25.10/node_modules/isomorphic-git/index.cjs:12665:16)
```

mojavelinux:
resolves #1933

This PR also exposes the Node.js globals related to compression to the Jest runtime so that these APIs get used if available. Without the code fix, the Node.js tests will now fail when using 22.3.0 with this modified Jest configuration.

### DIFF
--- a/__tests__/test-writeObject.js
+++ b/__tests__/test-writeObject.js
@@ -1,5 +1,5 @@
 /* eslint-env node, browser, jasmine */
-const { writeObject } = require('isomorphic-git')
+const { readObject, writeObject } = require('isomorphic-git')
 
 const { makeFixture } = require('./__helpers__/FixtureFS.js')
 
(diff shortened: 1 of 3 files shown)


jcubic:
As per @luantaraschi suggestion I'm closing this PR. The version 22.3.0 had a bug that was fixed.

PR #1948, opened 2024-07-03
fix: push tag: do not update remote ref

### LINKED ISSUE #1900: Pushing an tag creates a remote ref to an unexisting branch
I encountered a strange behavior when pushing an annotated tag. Upon investigation, I noticed that isomorphic-git's behavior is slightly different than Git's. I'm not sure it's a bug or may cause bugs down the line. I'm posting this question, to check with you if that might be the case.

### To reproduce
1. Clone an empty repository
2. Add a file, then git add and commit:  create a first commit on `main`
3. Push `main`
4. Create an annotated tag : `git tag -a v1.0 -m "add tag"`
5. Using isomorphic-git, push the tag
```
  const pushResult = await git.push({
    fs,
    http,
    dir: "<path to repo>",
    ref: "v1.0",
  });
```

**Result**
- On the Git server, the tag seems to be correctly pushed
- On my Git GUI (both VSCode's Git Graph and GitHub Desktop), a new "branch" is displayed called `origin/refs/tags/v1.0^{}` <= This is what ticked me off
image
- I looked at the `.git` folder, and isomorphic-git created a new ref at `refs/remotes/origin/refs/tags/v1.0`

**Comparing with git push**
Instead of calling `git.push()`, I ran the command `git push origin v1.0`

- On my Git GUI, I did not see a new "tag"
image
- Git did not modify the `.git` folder, and did not add a new ref

Do you have an explanation for this discrepancy? Does it have any impact down the line with future commands? Would you recommend that I investigate further isomorphic-git's code?

Notes
- I am using isomorphic-git 1.25.7 on Node
- My Git version is 2.41.0.windows.1

hhourani27:
fixes #1900

I added comments below on my changes. You feedback is welcome!

### DIFF
--- a/src/commands/push.js
+++ b/src/commands/push.js
@@ -283,7 +283,12 @@ export async function _push({
   }
 
   // Update the local copy of the remote ref
-  if (remote && result.ok && result.refs[fullRemoteRef].ok) {
+  if (
+    remote &&
+    result.ok &&
+    result.refs[fullRemoteRef].ok &&
+    !fullRef.startsWith('refs/tags')
+  ) {
     // TODO: I think this should actually be using a refspec transform rather than assuming 'refs/remotes/{remote}'
     const ref = `refs/remotes/${remote}/${fullRemoteRef.replace(
       'refs/heads',
--- a/__tests__/test-push.js
+++ b/__tests__/test-push.js
@@ -7,6 +7,7 @@ const {
   setConfig,
   push,
   listBranches,
+  resolveRef,
 } = require('isomorphic-git')
 
 const { makeFixture } = require('./__helpers__/FixtureFS.js')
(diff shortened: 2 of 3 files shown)


src/commands/push.js
```diff
@@ -283,7 +283,12 @@ export async function _push({
   }
 
   // Update the local copy of the remote ref
-  if (remote && result.ok && result.refs[fullRemoteRef].ok) {
+  if (
+    remote &&
+    result.ok &&
+    result.refs[fullRemoteRef].ok &&
+    !fullRef.startsWith('refs/tags')
```
hhourani27, line 290: I used the same condition used in line 222 to check if ref == tag

src/commands/push.js
```diff
@@ -283,7 +283,12 @@ export async function _push({
...
+  ) {
     // TODO: I think this should actually be using a refspec transform rather than assuming 'refs/remotes/{remote}'
     const ref = `refs/remotes/${remote}/${fullRemoteRef.replace(
       'refs/heads',
```
hhourani27, line 294: This line assumes that ref == branch. So we should make sure that ref != tag

jcubic:
Looks good, are there any tests that check if the push of a tag is updating refs?

hhourani27:
> Looks good, are there any tests that check if the push of a tag is updating refs?

Thanks. Pushing a tag should not update any ref in the local git repo.
I added a test on line 214 & 238 of `test-push.js` to make sure that the push tag is not creating a ref (reproducing the bug).
I also added a test on line 83 to make sure that pushing a branch does update the remote ref.

I'm open if you think we need more tests.

PR #1949, opened 2024-07-05
feat: commit --amend

### LINKED ISSUE #729: git commit --amend
I'm working on my new project using isomorphic git as history of save in editor. I'm in process of writing the app and I've added first commit but with wrong `author` It would be nice to have amend option that would modify existing commit. I can just trash whole git repo it just single commit nad new repo, but want to know if something like this exists in Isomorphic-git.

Or is there a way to have amend like functionally by editing files in .git directory?

hhourani27:
resolves #729 

## API changes
- Added a new parameter `amend` to `commit()`
- `message` parameter is no longer required if `amend === true`

## Rules applied when `amend ===  true` 
- Cannot amend if there's no commit (on an empty repo, before a first commit)
- If no `message` is provided, it takes the `message` of the commit pointed to by `ref`(or `HEAD` by default)
- By default it keeps the same `author` and `committer` information, unless they are overwritten in the method call

## Notes
- I significantly rewrote/rearranged `src/commands/commit.js`, so that adding `amend` does not seem so hacky. It's a major  command, so I added more tests to reassure myself that I did not break the commit functionality.
- I modified the jsdoc for `src/api/commit.js`, but I did not commit any changes to `website/`. I guess the website is updated during the github release process.
- I am trying to familiarize myself with the codebase. This is till now, my largest PR. Your feedback on my approach to implement this feature and on the code style is welcome!


jcubic:
Looks good.

PR #1907, opened 2024-05-03
fix: fix statusMatrix with win32 linebreaks

jayree:
## I'm fixing a bug or typo

fixes #1904

Issue #1904, opened 2024-04-26
StatusMatrix results are different on Windows and Linux


jayree:
I recently discovered that the results of statusMatrix are different on Windows and Linux.

`unmodified` files (`[1,1,1]`) are listed on windows as `modified, unstaged` (`[ 1, 2, 1 ]`)

See: https://github.com/jayree/iso-win-test/actions/runs/8846190106/job/24291562894

jcubic:
Hi, thanks for the report. Do you want to investigate why this is happening and maybe provide a fix? This project is community driven and don't have the main author that write all the code. There are a few maintainers, but we mostly manage the repository.

PR #1951, opened 2024-07-08
fix: don't access process without check

Vinzent03:
## I'm fixing a bug or typo

It took me a while to know why my [plugin](https://github.com/Vinzent03/obsidian-git) stopped working on mobile. This https://github.com/isomorphic-git/isomorphic-git/pull/1907/files#diff-3c52af97ec1834ad8729a4d3872436b50140f1bc6fe8e6732d234b7e983d6765R131 introduced the usage of `process` without any check of whether it even exists, since this package should be isomorphic.
I've now added a check to verify `process` exists.


jcubic:
Are you sure that `trustino` should be true for browser? Can you maybe create some unit test?

Vinzent03:
I'm unsure on how to create the correct unit test. But from reading the code, it should be working like before #1904 by setting `trustino` to true, doesn't it? It's the default value in `compareStats()` as well.

jcubic:
I will check this later. I'm not sure if LightingFS or BrowserFS use Inodes for their FS in status, that's why I was asking this.

jcubic:
I will merge this since it's better than throwing error.

jcubic:
I've checked lightingFS library, and it seems that the ino property in `fs.status` return unique value that identify the file.

PR #1958, opened 2024-10-07
refactor: use path.join from stdlib instead of custom join

scolladon:
## I'm fixing a bug or typo

This PR is the continuation of the #1861 one.
Which has been reverted because of the difference of implementation of the `path.join` depending the OS.

This patch aim to fix this compatibility issue by forcing the usage of `path/posix.join`


scolladon:
First issue : `path/posix` is available with webpack `5.65.0` meaning we need to upgrade it.
First try on my laptop fails with the configuration file format 👍 
FYI @mojavelinux

mojavelinux:
It fails even after the upgrade?

jcubic:
There is an error:

> Module not found: Error: Can't resolve 'path/posix' in '/home/vsts/work/1/s/src/utils'

jcubic:
When I use webpack in a different project, I use [path-browserify](https://www.npmjs.com/package/path-browserify)

scolladon:
> It fails even after the upgrade?

@mojavelinux I have not managed to upgrade properly because webpack changed the way the config file should be built.
I'm frustrated as well :) I'll try multiple things here and let you know

scolladon:
> When I use webpack in a different project, I use [path-browserify](https://www.npmjs.com/package/path-browserify)

I managed to make the compilation and the tests work on my machine, without having to upgrade webpack.
I assume it uses the posix implementation version everytime based on this [statement](https://github.com/browserify/path-browserify#api).
This is why I import it explicitely and not from the `path` rewrite `webpack` should do.

mojavelinux:
The problem is that this is just another implementation written in JavaScript. The whole point of the change is to take advantage of the native function provided in Node.js. If we just use a JavaScript implementation, we might as well avoid the dependency.

Is there no way to import one module for Node.js and a different module for webpack? I want it to use `path/posix` when running in Node.js.

scolladon:
I agree with you, it would be better if we could get the node version without dependencies.
The problem is this library is compatible with browser, hence the use of webpack.
I tried to use the import from `path/posix` but it failed at the webpack stage
When including just `path` webpack will use something very similar to (if not that itself) `path-browserify` behind the scene, which in the end is the [node v8 path implementation version](https://github.com/browserify/path-browserify/blob/872fec31a8bac7b9b43be0e54ef3037e0202c5fb/index.js#L1C4-L1C70)...

To me it looks like a smart tradeoff and will do what we want to achieve.

mojavelinux:
I just don't understand the point of this change if we aren't going to be able to use the native Node.js path/posix. Is path-browserify actually fixing bugs in the custom implementation being replaced? Because if it's not, having an extra dependency is worse.

mojavelinux:
> The problem is this library is compatible with browser

I'll admit that I've never understood why anyone would want to use this library in a browser. It's a cute trick, but not a realistic use case IMO. And I really dislike all the complexity that it adds to the project. So much so that I think that it might just be best for me to fork the project and strip it down to something that is focused on Node.js so I can avoid all the overhead and leave this project alone.

scolladon:
> I just don't understand the point of this change if we aren't going to be able to use the native Node.js path/posix. Is path-browserify actually fixing bugs in the custom implementation being replaced? Because if it's not, having an extra dependency is worse.

I don't think it fixes anything, its point is to make node js code available in the browser.
We could have used webpack 5 to use `path/posix` but in the end webpack would have change the real implementation with a configured polyfill.
Which is what we are currently doing by hand...

This dependency is not very good, I admit, but it is not that huge in term of size added, and it fixes the previous very slow implementation.

mojavelinux:
> it fixes the previous very slow implementation.

If it is faster, that's at least something. So that would warrant the extra dependency.

jcubic:
@mojavelinux should we merge this?

jcubic:
@scolladon Can you explain why some of the unit tests got replaced?

I also if this would be merged I need to check how much bigger the output file will be. Lets hope that path-browserify work with tree-shaking and not whole path is included.

scolladon:
> @scolladon Can you explain why some of the unit tests got replaced?

The unit test removed were there to test the gaps with the previous implementation and `path-browserify`.
It was important in the past but it does not seems to be the case anymore.
Now we are using `path-browserify` we could get rid of all the join test as well IMHO

> I also if this would be merged I need to check how much bigger the output file will be. Lets hope that path-browserify work with tree-shaking and not whole path is included.

It seems to be passing locally for me.

jcubic:
It seems that the output file is even smaller by 1KB.

PR #1996, opened 2024-12-23
Migrate from BrowserFS to ZenFS

### LINKED ISSUE #1910: BrowserFS deprecated
Hi there, I'm the maintainer of BrowserFS. I've been going though the dependents of browserfs.

Please see the notice in the readme: https://github.com/jvilk/browserfs

Would you be interested in updating your project to use the updated fork, [ZenFS](https://github.com/zen-fs/core)?

james-pre:
This PR migrates from using BrowserFS to ZenFS, since BrowserFS is deprecated.

This will resolve #1910


james-pre:
Tests should be updated to run with Node's current LTS (20) or the last, now deprecated LTS (18).

docs/fs.md
```diff
@@ -41,30 +41,40 @@ console.log(files);
...
+  }
+</script>
+<script type="module">
+import { fs, configureSingle } from "@zenfs/core";
+import { IndexedDB } from "@zenfs/dom";
+import git from "isomorphic-git";
+
+await configureSingle({ backend: IndexedDB });
+
+const files = git.listFiles({ fs dir: '/' });
```
jcubic, line 66: missing comma

src/api/status.js
```diff
@@ -106,16 +106,11 @@ export async function status({
         // If the oid in the index === working dir oid but stats differed update cache
         if (I && indexEntry.oid === workdirOid) {
           // and as long as our fs.stats aren't bad.
-          // size of -1 happens over a BrowserFS HTTP Backend that doesn't serve Content-Length headers
-          // (like the Karma webserver) because BrowserFS HTTP Backend uses HTTP HEAD requests to do fs.stat
-          if (stats.size !== -1) {
```
jcubic, line 111: You should not remove this code, the documentation may use ZenFS, but users may still use BrowserFS even that's deperecated. We don't want breaking changes.

src/models/GitWalkerFs.js
```diff
@@ -102,11 +98,7 @@ export class GitWalkerFs {
         const config = await GitConfigManager.get({ fs, gitdir })
         const autocrlf = await config.get('core.autocrlf')
         const content = await fs.read(`${dir}/${entry._fullpath}`, { autocrlf })
-        // workaround for a BrowserFS edge case
         entry._actualSize = content.length
-        if (entry._stat && entry._stat.size === -1) {
-          entry._stat.size = entry._actualSize
-        }
```
jcubic, line 109: The same please don't remove code for BrowserFS compatibility from the codebase, we will need to release version 2.0 for no reason.

jcubic (changes requested):
Please remove the changes made to the Codebase.

src/models/GitWalkerFs.js
james-pre: Got it.

jcubic:
It seems that a change in Node.js version break everything. Because it now require extensions inside import

The error looks like this:

```
ERROR in ./src/utils/unionOfIterators.js 1:0-58
Module not found: Error: Can't resolve '../models/RunningMinimum' in '/home/vsts/work/1/s/src/utils'
Did you mean 'RunningMinimum.js'?
BREAKING CHANGE: The request '../models/RunningMinimum' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
```

src/models/GitWalkerFs.js
james-pre: https://github.com/isomorphic-git/isomorphic-git/pull/1996/commits/7ccd61703d9a2a85c04d32f5320752e11fc215e7

src/api/status.js
james-pre: https://github.com/isomorphic-git/isomorphic-git/pull/1996/commits/7ccd61703d9a2a85c04d32f5320752e11fc215e7

docs/fs.md
james-pre: https://github.com/isomorphic-git/isomorphic-git/pull/1996/commits/eb53f68a289e12ffca54baf91654700e689477a6

james-pre:
> It seems that a change in Node.js version break everything. Because it now require extensions inside import

I just went through and added all of the missing extensions =)

> The documentation may use ZenFS, but users may still use BrowserFS even that's deperecated. We don't want breaking changes.

That makes sense, I've gone ahead and re-added support


Due to an update of Webpack and some related dependencies, this PR reduces the number of dependencies by about 300 and decreases the number of vulnerabilities from 86 (6 low, 26 moderate, 44 high, 10 critical) to 61 vulnerabilities (2 low, 20 moderate, 32 high, 7 critical).

jcubic:
I'm not sure if this was the right move, now there errors about `require`:

```
Chrome 128.0.0.0 (Android 10) ERROR
  An error was thrown in afterAll
  Uncaught ReferenceError: require is not defined
  ReferenceError: require is not defined
      at ./__tests__/index.webpack.js (/tmp/_karma_webpack_966385/commons.js:12:1)
      at __webpack_require__ (/tmp/_karma_webpack_966385/runtime.js:23:41)
      at __webpack_exec__ (index.webpack.1694758118.js:4:48)
      at index.webpack.1694758118.js:5:55
      at __webpack_require__.O (/tmp/_karma_webpack_966385/runtime.js:57:23)
      at index.webpack.1694758118.js:6:56
      at webpackJsonpCallback (/tmp/_karma_webpack_966385/runtime.js:116:39)
      at index.webpack.1694758118.js:2:81
```

james-pre:
I'll take a look at it when I can. Merry Christmas!

jcubic:
There is a need to migrate to ZenFS, since it may help with:
* https://github.com/isomorphic-git/isomorphic-git/pull/2090

Right now, the tests are disabled for Browser, migration to ZenFS may fix the issue with running tests in browser.

I will try to merge with main locally.

jcubic:
Your last commit didn't trigger the build, it's hanging, and you can't click to see the status.

I will try to check again tomorrow, to trigger the build or check Azure DevOps if I can find the status of your PR.

jcubic:
This type got an error:

```
__tests__/__helpers__/FixtureFS.js(5,9): error TS2459: Module '"./FixtureFS/makeZenFS.js"' declares 'makeZenFS' locally, but it is not exported.
```

It looks like the timeout error is gone, probably some temporary glitch outside of our code.

Can you fix this type error?

james-pre:
Hey @jcubic, the error was because the version of Eslint and Standard was too old.

Summary of the last few commits:
- Updated Eslint and Standard, including some semi-automatic fixes
- Updated Jest and the Jest plugins
	- This also included updating the format of some snapshots
- Changed the tests (including the helpers) to use ESM instead of CJS
- Changed src/utils/join.js to use `@zenfs/core/path` instead of `path-browserify` since `path-browserify` doesn't support ESM for some reason
- Fixed the command used to generate the index.json command

With all of that, I'm getting these results when running tests:

```
Test Suites: 4 failed, 83 passed, 87 total
Tests:       11 failed, 5 skipped, 578 passed, 594 total
Snapshots:   95 passed, 95 total
Time:        36.794 s, estimated 37 s
```

Apart from some timeouts that appear unrelated to the changes, there is only a *single* failure:
```
 FAIL  __tests__/test-packObjects.js
  ● Test suite failed to run

    SyntaxError: The requested module 'isomorphic-git/internal-apis' does not provide an export named 'readObject'
```

I'm not sure what is causing this but it also seems unrelated.

jcubic:
bundlewatch says that the code is 15.8% bigger. Probably because with path module from Zenfs.

james-pre:
> Probably because with path module from Zenfs.

I did some testing, it looks like that does add about 4KB, but I'm not sure where the other 7KB are coming from. I've gone ahead and inlined the bare minimum need for `path.join`, which cuts the bundle size by that 4KB.

The other 7KB come from changes before 4014b7dabfe091e6736bd4c18d6cb882a721758e. I would think a lot of it comes from adding `.js`, but I have no idea.

james-pre:
I went ahead and diff'ed `src` excluding changes to import/export specifiers:
```sh
git diff -I "(import|export) .* from '.*'" main..migrate-zenfs src
```

The result was the inlining of `path.join` along with some semi-automatic changes to fix Eslint issues:
- 8 lines changed to use shorthand [property definitions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#property_definitions) (automatic)
- 1 line changed to use `Array.prototype.forEach` instead of `Array.prototype.map`
- 2 lines changed to remove redundant/unused `return`
- 1 line changed to use a regex literal instead of `new RegExp`

Statistics excluding import/export specifier changes:
```
 src/api/updateIndex.js                        |  2 +-
 src/commands/push.js                          |  2 +-
 src/commands/walk.js                          |  2 +-
 src/http/web/index.js                         |  2 +-
 src/models/FileSystem.js                      |  2 --
 src/models/GitIndex.js                        |  2 +-
 src/models/GitWalkerFs.js                     |  2 +-
 src/utils/abbreviateRef.js                    |  2 +-
 src/utils/flatFileListToDirectoryStructure.js |  2 +-
 src/utils/join.js                             | 96 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++-
 src/utils/parseAuthor.js                      |  4 ++--
 11 files changed, 105 insertions(+), 13 deletions(-)
```

I also counted the number of bytes changed with `gawk` using [this script](https://github.com/user-attachments/files/22349332/count-changed-bytes.sh):
- 2739 bytes added, 367 bytes deleted, net +2372 when import/export specifier changes
- 6964 bytes added, 4319 bytes deleted, net +2645 when including specifier changes

While this seems like an increase, we get rid of 54.3 kB by dropping dependence on `path-browserify`.

So, the majority of the bundle size change probably comes from dependencies.

jcubic:
Right now there is syntax error in shell script:

```
nps is executing `test.jest` : cross-env-shell NODE_OPTIONS=--experimental-vm-modules (timeout -t 5m -- jest --ci --coverage) || (timeout -t 5m -- jest --ci --coverage) || (timeout -t 5m -- jest --ci --coverage)
/bin/sh: 1: Syntax error: "(" unexpected
```

Maybe it should be executed by bash. Did you change anything related to this code?

jcubic:
Now we finally get back to those timeout errors. I have no idea what they are and why they are happening. The only thing that I can come up with is the CORS proxy, but I'm not sure if it's used in Unit test since this is Node.js.

Maybe increasing timeout will help.

james-pre:
It looks like all of the timeouts occur due to a problem with the local CORS proxy.

james-pre:
Hey @jcubic,

I was looking to the CORS proxy and ended up rewriting it to use ESM and use built-ins (https://github.com/james-pre/cors-proxy/tree/fix). Surpringly this almost works though GitHub rejects the requests (`Please upgrade your git client.
GitHub.com no longer supports git over dumb-http: https://github.com/blog/809-git-dumb-http-transport-to-be-turned-off-in-90-days`)

> But it seems that you can't run test locally anymore. Every single require return an error

This is because `require` is CJS-only. Node.js supports `import`ing CJS, however I don't think this works with the VM modules used by Jest. 

> It looks like your changes changed dependencies of Cors-Proxy, and it doesn't work anymore.

I did a `git blame` and it looks like the change to package-lock.json was made in ad4350ab276a4926fdcf78a130365753d65c0e56:

```
$ git diff ad4350a~1..ad4350a package-lock.json | grep -A 2 \"node_modules/micro\"
+    "node_modules/micro": {
+      "version": "9.4.1",
+      "resolved": "https://registry.npmjs.org/micro/-/micro-9.4.1.tgz",
--
-    "node_modules/micro": {
-      "version": "9.3.4",
-      "resolved": "https://registry.npmjs.org/micro/-/micro-9.3.4.tgz",
```

> now there is this error, and few timeouts but less than before.

So that error is cause because the `readObject` function required here doesn't exist:
```
const { readObject, readObjectPacked } = require('isomorphic-git/internal-apis')
```

When moving to ESM though, the module loader actually detects this.

jcubic:
The merge commit in blame shows that it was introduced in this commit because there was the conflict. The conflict was in package.json or package-lock.json that's why the error don't appear in main where the old version of micro.

jcubic:
Sorry, I didn't test the cors-proxy that you mention.

Do you want to contribute to: https://github.com/isomorphic-git/cors-proxy/ but there are a lot of changes on top of the upstream repo? I'm not sure if you can easily apply that patch.

jcubic:
CORS proxy v3.0.0 finally got released. We can move forward.

jcubic:
There are a few errors:

```
Unhandled promise rejection: Error: Cannot find module 'fs'
```

```
SyntaxError: Unexpected token '.'
```

I don't see the location.

This dozens of errors like this:
```
export 'writeTree' (imported as 'writeTree') was not found in 'isomorphic-git'
```

Did you update any dependencies from last time the CI/CD run.

I'm attaching the logs:

[Azure.log](https://github.com/user-attachments/files/22555711/Azure.log)

jcubic:
And why did you delete `is-git-ref-name-valid` package, the code is the same. You also copy/pasted the code without adding copyright and attribution.

jcubic:
Ok, will you be able to make the CI/CD test pass? I hope that this whole dependency happen only in test file, that somehow are bundled with Webpack and not the main library.

As for `is-git-ref-name-valid`, I was also concerned about any tampering. I've added the repo to the watch list. And I looked closely at the PR author and the author of the library. 3.5kB that happen only in node_module is not an issue, if the bundle is not bigger.

Also note that if the code is included, the whole LICENSE file need to be merged with our license file. And there is need to be clear indication from where the code come from. It's simpler to just include a dependency, that as you said it's only 250 bytes in the output bundle.

james-pre:
> Ok, will you be able to make the CI/CD test pass? I hope that this whole dependency happen only in test file, that somehow are bundled with Webpack and not the main library.

I think a lot of the problem was Karma being so old. In 5ef9081a380e75704a6df598aa36a32c86f22560 I replaced Karma with jest-puppeteer, which fixes *all* of the problems.

> As for `is-git-ref-name-valid`, I was also concerned about any tampering. I've added the repo to the watch list. And I looked closely at the PR author and the author of the library. 3.5kB that happen only in node_modules is not an issue, if the bundle is not bigger.
> 
> Also note that if the code is included, the whole LICENSE file need to be merged with our license file. And there is need to be clear indication from where the code come from. It's simpler to just include a dependency, that as you said it's only 250 bytes in the output bundle.

Most people use npm, which means they will still pull in all 3.5kB. isomorphic-git has [50 direct and transitive dependencies](https://npmgraph.js.org/?q=isomorphic-git), yet the LICENSE.txt file included with the bundle only includes 4 of them. We include a reference to the license (`SPDX-License-Identifier: MIT`), which should be sufficient.

.github/workflows/ci.yml
```diff
@@ -100,73 +101,6 @@ jobs:
           name: index.umd.min.js
           path: index.umd.min.js
 
-  test-in-browser:
```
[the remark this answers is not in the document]
james-pre: I moved the browser testing into the main unit tests job since it's now handled with a `nps` script (`test.chrome`, more browsers can be added if puppeteer supports them)

.all-contributorsrc
```diff
@@ -766,6 +766,88 @@
...
+    {
+      "login": "acandoo",
+      "name": "acandoo",
+      "avatar_url": "https://avatars.githubusercontent.com/u/117209328?v=4",
+      "profile": "https://github.com/acandoo",
+      "contributions": [
+        "platform",
+        "userTesting"
+      ]
+    },
```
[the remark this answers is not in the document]
jcubic: ok, they appear on the main branch, so I think that we can ignore it.

__tests__/__helpers__/FixtureFS/makeNodeFixture.js
```diff
@@ -1,29 +1,49 @@
...
+
+  if (fixturePath) {
+    await _fs.promises.cp(fixturePath, tempDir, { recursive: true })
+  }
 
-  const {
-    getFixturePath,
-    createTempDir,
-    copyFixtureIntoTempDir,
-  } = require('@wmhilton/jest-fixtures')
```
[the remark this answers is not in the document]
jcubic: Sure, you can remove it.

jcubic:
About the browser testing, BrowserStack and SauceLabs are probably way better than puppeteer. They run on iOS, Android and Safari. You can't test the same with puppeteer.

james-pre:
We aready had puppeteer so I just went with that— I'm looking into other libraries/frameworks. AFAIK BrowserStack and SauceLabs are just launchers and don't provide an API for running tests. Playwright and Vitest look promising.

jcubic:
BrowserStack and SauseLabs main purpose is to run automated tests from CI/CD. This is what Azure DevOps is doing.

jcubic:
Just merged a PR with new tests for git stash, you need to merge with main and resolve conflicts.

If you want to rewrite the test with vitest, I suggest doing this in different PR. Please restore back GitHub workflow so all automated tests run. I then will review the rest of the code, and we will be able to finally merge.

Let's limit the scope to what is really needed here.

jcubic:
Yes, they need to be fixed in a different PR, but not by deleting everything.

james-pre:
@jcubic I completely reverted the workflow except for changing the Node.js version.

azure-pipelines.yml
```diff
@@ -15,15 +15,9 @@ jobs:
...
         inputs:
-          versionSpec: '14.x'
+          versionSpec: '22.x'
         displayName: 'Install Node.js'
 
-      - task: 1ESLighthouseEng.PipelineArtifactCaching.RestoreAndSaveCacheV1.RestoreAndSaveCache@1
-        inputs:
-          keyfile: 'package-lock.json'
-          targetfolder: 'node_modules'
-          vstsFeed: $(ARTIFACT_FEED)
```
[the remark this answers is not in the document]
james-pre: IIRC that job was doing browser testing but with the move to puppeteer the browser testing is done in the main job.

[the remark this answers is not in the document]
james-pre: @jcubic I fixed the current release issue in #2238.

PR #2023, opened 2025-01-16
perf: cache git config (statusMatrix ~20% faster)

### LINKED ISSUE #2022: perf: cache git config
Improve performance of status matrix by cache the git config in the GitConfigManager.
This avoids re reading the config repeatedly.

lukecotter:
resolves #2022

In my testing these changes improved performance of status matrix in my large repo by ~20%


src/managers/GitConfigManager.js
```diff
@@ -2,17 +2,24 @@ import { GitConfig } from '../models/GitConfig.js'
...
+    )
+    return this.config
   }
 
   static async save({ fs, gitdir, config }) {
     // We can improve efficiency later if needed.
     // TODO: handle saving to the correct global/user/repo location
+    this.config = config
     await fs.write(`${gitdir}/config`, config.toString(), {
-      encoding: 'utf8',
```
jcubic, line 15: Please don't change the style the library is using.

src/managers/GitIndexManager.js
```diff
@@ -20,8 +20,11 @@ function createCache() {
 }
 
 async function updateCachedIndexFile(fs, filepath, cache) {
-  const stat = await fs.lstat(filepath)
-  const rawIndexFile = await fs.read(filepath)
+  const [stat, rawIndexFile] = await Promise.all([
+    fs.lstat(filepath),
+    fs.read(filepath)
+  ])
```
jcubic, line 26: Please don't do usless refactor

jcubic (changes requested):
the tests are failing because of ESLint (missing dangling comma)

src/managers/GitIndexManager.js
```diff
@@ -20,8 +20,11 @@ function createCache() {
 }
 
 async function updateCachedIndexFile(fs, filepath, cache) {
-  const stat = await fs.lstat(filepath)
-  const rawIndexFile = await fs.read(filepath)
+  const [stat, rawIndexFile] = await Promise.all([
+    fs.lstat(filepath),
+    fs.read(filepath)
```
jcubic, line 25: missing dangling comma

src/managers/GitConfigManager.js
lukecotter: Sorry about that, I am not sure what happened there I must have something misconfigured.
I have run the formatting script now.

src/managers/GitIndexManager.js
lukecotter: done

jcubic:
The tests are failing

lukecotter:
> The tests are failing

I will get that test resolved asap! 👍🏻

lukecotter:
@jcubic 
I have the jest test passing now. 
The browser tests are failing, it seems to be because the config cache persists between test cases which does not happen in the node version. Each of the browser tests creates a new in memory file system and new git config file.

With the current cache implementation test 1, file 1 creates a new config file in the in memory fs and adds to the config cache. Test 2, file 1 creates a new config in the in memory fs. Previously the gitConfig file would have been re read but now the cached version  from test 1 comes back, which result in values in the config from test 1 come back. 

There are a number of ways this can be resolved.

It is possible in practice that this is a non issue and there is a config in the browser tests that can be changed to ensure test isolation. I am not too familiar with karma and this browser test setup however.

I could instead add a check on mtime to determine if the cache file has changed and then invalidate the cache if it has. I worry that this will negate the performance gains of avoiding re reading the file, but I am yet to benchmark this. 

There is a final approach. After a more careful review of the code it seems there are only a few areas of the code where the gitConfig is read in situations where repeatedly re reading is a performance concern, this is mainly in loops.
I can instead create a much smaller and localised cache for these situations, gitWalkerFs is one of these. 

I am favouring this last approach as it is safer and will still have the performance gains. If you have any thoughts let me know.

jcubic:
I don't like your solutions at all since they require writing code just for the tests. Instead I would add a way to invalidate the cache manually and clear the cache before each test that read/write into Config (in `beforeEach` or `afterEach`).

lukecotter:
> I don't like your solutions at all since they require writing code just for the tests. Instead I would add a way to invalidate the cache manually and clear the cache before each test that read/write into Config (in `beforeEach` or `afterEach`).

I switched to a more narrow cache of the config in [GitWalkerFs.js](https://github.com/isomorphic-git/isomorphic-git/pull/2023/files?diff=unified&w=1#diff-3c52af97ec1834ad8729a4d3872436b50140f1bc6fe8e6732d234b7e983d6765) before I saw your comment. All tests are now passing. 

If you would still prefer me to change back to a cache in [GitConfigManager.js](https://github.com/isomorphic-git/isomorphic-git/blob/main/src/managers/GitConfigManager.js) and add a funnction to clear the cache between tests I am happy to do that too 😄

jcubic:
I think it's fine. Can you add some new tests? Especially reading from config, writing to config, and reading again? If this doesn't work, you will need to add manual cache invalidation. Or if it requires extra step, a documention update.

src/models/GitWalkerFs.js
```diff
@@ -115,6 +117,7 @@ export class GitWalkerFs {
 
   async oid(entry) {
     if (entry._oid === false) {
+      const me = this
```
jcubic, line 120: I would use a common `self` or `that` instead of `me`.

jcubic (changes requested):
One small thing.

PS: I suggest that you do a refactor in separated PR of the code that is unrelated to the feature you're implementing or bug you're fixing.

src/models/GitWalkerFs.js
lukecotter: done

lukecotter:
> One small thing.
> 
> PS: I suggest that you do a refactor in separated PR of the code that is unrelated to the feature you're implementing or bug you're fixing.

I have moved the improvements / refactor to this [PR](https://github.com/isomorphic-git/isomorphic-git/pull/2063).
Now that I re read your comment I realise you may have meant in future.

PR #2035, opened 2025-02-06
fix: Use default import and destructure method because `path-browserify` uses `module.exports`

### LINKED ISSUE #2004: Explicit use of path-browserify breaks isomorphic-git on Deno
The explicit use of path-browserify (1.27.2, #1958) instead of using `'path'` and letting the bundler polyfill has the consequence of breaking isomorphic-git on Deno:

```
error: Uncaught SyntaxError: The requested module 'path-browserify' does not provide an export named 'join'
    at <anonymous> (file:///Users/michael/Library/Caches/deno/npm/registry.npmjs.org/isomorphic-git/1.27.3/index.js:3:10)
```

Before this change, `isomorphic-git` 1.27.1 was working fine in Deno when installed through Deno's NPM support (which supplies its own version of the Node `path` package).

eliot-akira:
This resolves #2004.

The issue was that `path-browserify` uses `module.exports` to export only `default`, but on the side of `isomorphic-git` ([src/utils/join.js](https://github.com/isomorphic-git/isomorphic-git/blob/d9283029518d63bc7b3f32cfedc0616f26dc1608/src/utils/join.js)) it was using named exports, which throws an error with Deno and possibly other bundlers.

The solution is to import `default` and destructure the needed method.

## I'm fixing a bug or typo


jcubic:
This load whole path library, it's not a solution. But if you want to fork the path-browserify and refactor it to be able to use it as ES Module, we can put into into the org. You can create a fork and after done transfer ownership to isomorphic-git. I'm not sure if you need to be part of the org.

We can also publish it in @isomorphic-git or on npm.

eliot-akira:
From a [search on NPM](https://www.npmjs.com/search?q=path-browserify), there are two forks that have converted the library to ESM that are close but not quite acceptable.

- https://github.com/bundled-es-modules/path-browserify/blob/main/index-esm.js

  This one is not tree-shakeable, and tries to polyfill `process`. It has some questionable wrapper to support CommonJS/ESM.

- https://github.com/Hocti/path-browserify-esm/blob/master/index.esm.js

  Also not tree-shakeable because all methods are defined as properties of the `posix` object. Otherwise it might be a good reference for creating a fork specifically for the purposes of `isomorphic-git`.

---

> You can create a fork and after done transfer ownership

Sure, that makes sense to keep it under the `@isomorphic-git` namespace, so other people can work on it in the future if needed.

OK, I'll come back with a fork soon and update this PR to use it when it's ready.

eliot-akira:
OK, I've prepared a fork converted to ES Module, which exports its methods individually to support tree shaking. All original tests pass.

- https://github.com/eliot-akira/path-browserify

I learned that I'm not able to transfer it to `isomorphic-git` unless I have the ability to create a new repo in the org.

> To transfer a repository that you own to an organization, you must have permission to create a repository in the target organization.

(From [GitHub Docs: Transferring a repository](https://docs.github.com/en/repositories/creating-and-managing-repositories/transferring-a-repository))

If you'd like, you can give me the permission temporarily until the transfer is complete. After that you can publish the package under a new name, then I can continue finishing up this pull request.

jcubic:
I've send an invitation.

eliot-akira:
Perfect, I completed the transfer.

- https://github.com/isomorphic-git/path-browserify

jcubic:
Great. You can try to install using npm from github url, to see if it will work with Deno, and if bundlewatch don't report big increase in size.

Also it would be nice to update the README to show how to use the library. You can use `@isomorphic-git/path-browserify` npm package. If everything will work I will publish it to npm.

I can also publish tests versions, if the code work. This will be first time I publish manually npm package to `@isomorphic-git` scope, I hope that I have the permission.

eliot-akira:
> try to install using npm from github url, to see if it will work with Deno, and if bundlewatch don't report big increase in size.

Good idea, I'll see how it goes. I'm not a regular user of Deno (but I do use Bun which is similar), so I'm curious if the original issue with import will be resolved.

> would be nice to update the README

That's true, the information in the README is outdated now, especially with changes after the fork. I'll start a pull request in that repo.

jcubic:
The package [@isomorphic-git/path-browserify](https://www.npmjs.com/package/@isomorphic-git/path-browserify] got published.

jcubic:
You added the file to gitignore:

https://github.com/isomorphic-git/path-browserify/blob/main/.gitignore

jcubic:
published a new version

eliot-akira:
Two issues now:

- `index.umd.min.js` +4.28KB
- Jest fails to parse ES Module in `node_modules`

eliot-akira:
I've created a separate issue for the `ERR_OSSL_EVP_UNSUPPORTED` error, which is unrelated to this PR.

- #2047

---

For the unexplained bundle size increase, it seems to be an issue with `bundlewatch` itself.

With a clean clone of this repo, the build results in:

```
PASS ./index.umd.min.js: 77.36KB < 100KB (gzip)

bundlewatch PASS
Everything is in check (+77.36KB, -0B)
```

This is actually bigger than the result of this PR, which is `77.13KB`. So the report of `+4.28KB` is incorrect, it should be `-0.23KB`.

---

I'm no longer seeing the Jest parse error, so the last issue I'm still solving is:

- `test.typecheck` fails due to type checking modules in `node_modules/@types` such as `babel__traverse`

eliot-akira:
For the first of last two remaining issues:

- `test.typecheck` fails due to type checking modules in `node_modules/@types` such as `babel__traverse`

I found relevant info here:

- [Angular: node_modules/@types/babel _traverse/index.d.ts(1137,43): error TS1109: Expression expected](https://stackoverflow.com/questions/63903376/angular-node-modules-types-babel-traverse-index-d-ts1137-43-error-ts1109)

It's caused by a descendent dependency `@types/babel__traverse` being declared with version "*", which downloads the newest type declaration file `index.d.ts`, whose syntax is not compatible with the old version of TypeScript used in this project (`3.9.0-dev.20200223`).

A workaround suggested in the article works, to override it to a specific version.

```json
  "overrides": {
    "@types/babel__traverse": "7.0.6"
  },
```

With this, the errors about `babel__traverse` are gone; however, there are two remaining errors from another package.

```
nps is executing `test.typecheck` : tsc -p tsconfig.json
node_modules/@types/gensync/index.d.ts:52:34 - error TS1005: ',' expected.

52         errback(...args: [...args: A, callback: Callback<R, E>]): void;
                                    ~

node_modules/@types/gensync/index.d.ts:52:47 - error TS1005: ',' expected.

52         errback(...args: [...args: A, callback: Callback<R, E>]): void;
                                                 ~
Found 2 errors.
```

I tried to override this package [`@types/gensync`](https://www.npmjs.com/package/@types/gensync?activeTab=versions) also, but even with the oldest version available `1.0.0`, the error is not resolved.

The real solution is to update `typescript` to a newer version that supports the new type declaration syntax. The newest is `5.7.3`, which is two major versions ahead so there may be unintended side effects.

jcubic:
The issue say that the experimental flag will be removed in Node 12.

Maybe updating Node.js to something more recent will fix the issue. I'm not sure what Node is used. Node 18 is the last supported version, but only with security fixes. So maybe drop everything below 18.

I would update jest in a separated PR.

eliot-akira:
> The issue say that the experimental flag will be removed in Node 12.

They say "*ESM support* will be unflagged in a future release of Node 12". However, Jest uses the `vm` API ([https://nodejs.org/api/vm.htm](https://nodejs.org/api/vm.html)), which does not yet support ESM without the experimental flag. Here's the newest comment (Jan 2025) in Node's repo:

> It looks like this (enable `--experimental-vm-modules` by default) didn't happen in Node 23.
>
> Maybe it can happen in Node 24 instead?

- https://github.com/nodejs/node/issues/37648#issuecomment-2568320055

So I guess the use of the flag is not problematic, since they're planning to turn it on by default in a future Node version release.

> updating Node.js to something more recent

Locally I'm using Node v22, and Jest still doesn't work without the flag.

> update jest in a separated PR

The difficulty is that Jest works fine with the original version of `path-browserify` (CommonJS), so I won't be able to see how it will work with the new fork (ESM) if I work on it in a separate PR.

Well, I'll try and see.

eliot-akira:
Thanks!

PR #2053, opened 2025-03-08
Remove first person from README.md

louwers:
As the README.md mentions, this is now a community maintained project.

It is unclear who the "I" and "me" in the README.md refers to. Therefore I changed it to the third person.

Also fixes the link to Bill's new GitHub handle.


louwers:
I noticed that the docs sometimes also use first person.

PR #2069, opened 2025-04-22
chore: Run tests in GitHub Workflow

seanpoulter:
In order to resolve and merge #1902 I want to rerun the GitHub Workflow and start with just one version of Node.js.

### Summary

* Add a GitHub Workflow that runs on a pull_request and on demand (workflow dispatch)
* Run the tests using Node.js v14
* Use `xvfb` (X Virtual Frame Buffer) to run the browsers without a display (headless)

### Next Steps

If we can configure the GitHub Workflow to pass, we'll continue adding the Node.js versions that are long-term support (LTS) releases (16, 18, 20, 22).


seanpoulter:
I believe this pull request is failing in the pipeline because it does not have access to the secrets. I'll create a branch in this repo and try again. 🤞

PR #2050, opened 2025-02-25
Upgrade TypeScript to newest version

eliot-akira:
This pull request updates the dependency `typescript` to the newest version `5.7.3`.

It resolves the type checking issues mentioned in [#2035](https://github.com/isomorphic-git/isomorphic-git/pull/2035#issuecomment-2677593012), where new declaration files for some dependencies were not able to be parsed correctly by an old version of TypeScript.

> It's caused by a descendent dependency `@types/babel__traverse` being declared with version *, which downloads the newest type declaration file `index.d.ts`, whose syntax is not compatible with the old version of TypeScript used in this project (`3.9.0-dev.20200223`).

---

It also resolves new type errors that resulted from the update.

```
nps is executing `test.typecheck` : tsc -p tsconfig.json
__tests__/__helpers__/find-unused-errors.cjs:1:9 - error TS2614: Module '"../.."' has no exported member 'E'. Did you mean to use 'import E from "../.."' instead?

1 const { E } = require('../..')
          ~

__tests__/__helpers__/find-unused-errors.cjs:9:23 - error TS2349: This expression is not callable.
  Type 'typeof import("replace-in-file")' has no call signatures.

9     let files = await replace({
                        ~~~~~~~

__tests__/__helpers__/karma-successful-browsers-reporter.cjs:27:12 - error TS2339: Property 'failedBrowsers' does not exist on type 'onBrowserComplete'.

27       this.failedBrowsers.push(browser)
              ~~~~~~~~~~~~~~

__tests__/__helpers__/karma-successful-browsers-reporter.cjs:29:12 - error TS2339: Property 'successfulBrowsersFullNames' does not exist on type 'onBrowserComplete'.

29       this.successfulBrowsersFullNames.push('X ' + browser.name)
              ~~~~~~~~~~~~~~~~~~~~~~~~~~~

__tests__/__helpers__/karma-successful-browsers-reporter.cjs:31:12 - error TS2339: Property 'successfulBrowsersFullNames' does not exist on type 'onBrowserComplete'.

31       this.successfulBrowsersFullNames.push(browser.name)
              ~~~~~~~~~~~~~~~~~~~~~~~~~~~

__tests__/test-addRemote.js:29:50 - error TS2322: Type 'undefined' is not assignable to type 'string'.

29       await addRemote({ fs, dir, gitdir, remote, url })
                                                    ~~~

  index.d.ts:889:5
    889     url: string;
            ~~~
[...]

PR #2076, opened 2025-04-24
Convert tests from CommonJS to ES Module syntax

eliot-akira:
This pull request converts the tests from CommonJS to ES Module syntax.

> CommonJS can be recognized by the use of the `require()` function and `module.exports`, while ES modules use import and export statements 

It's part of a set of changes that created a fork of `path-browserify` with modernized codebase and tree-shakeable exports (ESM); upgrade TypeScript in the build setup and supported browser matrix; updating Jest config and convert everything including tests to ES Module so the new fork can be tested together.

- #2035 
- #2050 

After this PR is refined and merged, #2035 should fast-forward/merge from main branch to complete its final step.


eliot-akira:
Next to be solved, this use of `await import('fs')`.

```
ReferenceError: You are trying to `import` a file after the Jest environment has been torn down. From __tests__/test-log-file.js.

       7 |
       8 | export async function makeNodeFixture(fixture) {
    >  9 |   const _fs = Object.assign({}, await import('fs'))
         |                                 ^
      10 |
      11 |   const fs = new FileSystem(_fs)
      12 |

      at makeNodeFixture (__tests__/__helpers__/FixtureFS/makeNodeFixture.js:9:33)
      at makeNodeFixture (__tests__/__helpers__/FixtureFS.js:10:54)
      at Object.makeFixture (__tests__/test-readObject.js:119:34)
```

A similar error elsewhere.

```
Error [ERR_VM_MODULE_NOT_MODULE]: Provided module is not an instance of Module

       7 |
       8 | export async function makeNodeFixture(fixture) {
    >  9 |   const _fs = Object.assign({}, await import('fs'))
         |                                 ^
      10 |
      11 |   const fs = new FileSystem(_fs)
      12 |

      at makeNodeFixture (__tests__/__helpers__/FixtureFS/makeNodeFixture.js:9:33)
      at Object.<anonymous> (__tests__/test-push.js:21:28)
```

https://dev.azure.com/isomorphic-git/isomorphic-git/_build/results?buildId=3646&view=logs&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=5b4cc83a-7bb0-5664-5bb1-588f7e4dc05b

Maybe `(await import('fs')).default` is necessary.

Weirdly I'm not seeing these errors when I run the tests locally. It may depend on Node.js version.

jcubic:
I can't merge when there are failing tests. If you want to work on this when a PR may break things, I can create a branch for you that you will merge stuff into. When the code will be ready, we will merge that branch as one thing. Let me know if this will work.

eliot-akira:
Sure, understood. I'm still working through the errors, I'll see if they can be completely solved in this PR by itself while still using the original `path-browserify`.

lemanschik:
@eliot-akira do not forget we can always Polyfill backward later eg supply URL but i vote for node 18 anyway as smallest target even for tests and we need to get them passing with node 18 anyway. But Readme States that this then needs version bumb to v2 as it is breaking compatibility matrix but that needs to get done anyway even out of security view this needs to get tested on newst chrome and then we later supply backward compat.

jcubic:
I don't think that we need to bump major version after dropping Node. We only need to update everywhere that we don't support them anymore.

lemanschik:
@jcubic good point in real world we would just do 1.xx+1 as we do not real break the existing API and only do additions. But we should also update the readme with that update as it states clear as soon as version matrix of support is broken we release major 1+1.xxxx

eliot-akira:
Yes, this PR modifies **all tests** to convert them to ES Modules. This would be necessary sooner or later for the health of this codebase. As for the PR mentioned:

- #2090 

Well, there's no getting around it: if that's merged first, I will have to sync this branch with `main` again and make any necessary updates. That's true for any other PR that might get merged before this one.

eliot-akira:
Well, no problem, it's just a matter of resolving differences. I don't mind either way, if the other PR #2090 gets merged before or after this #2076.

But if I try and look at it objectively, I think this one makes sense to merge first. It **doesn't add or change any functionality**. In fact the point is to perfectly maintain the same behavior as before.

In contrast, the other PR adds a big feature with many tests to support it (which is great and looks useful). It's worth taking the time to review and release.

I don't have the script I wrote to convert `require()` to `import` across all tests. It was a one-liner command piping the result of `grep` (matching files) into a `while` loop to apply `sed` to replace patterns. It missed some multi-line `require` statements, which I updated manually.

jcubic:
Just checked the logs, there is a flood of errors:

> SyntaxError: Cannot use 'import.meta' outside a module

james-pre:
@eliot-akira Please rebase from main— I actually converted the tests to use ESM in #1996, so this PR might not even be needed. We still need to run tests on more platforms though so that is always possible.

eliot-akira:
> I actually converted the tests to use ESM in https://github.com/isomorphic-git/isomorphic-git/pull/1996

Great, from what I see, that PR removes the need for these related PRs, which can be closed:

- #2076
- #2035

I think these were better to do together in one big effort, as you did in #1996 (which I notice started much earlier). So the following issues are solved and can also be closed.

- #2004
- #2117

Nice work! Solving all these in one swift cut. [ZenFS](https://zenfs.dev/core/) looks good too, for future extensibility.

lemanschik:
this should be closed.

jcubic:
@eliot-akira closing the PR. If you decide that there is still something to do here, open a new PR.

PR #2079, opened 2025-04-25
chore: Update bundlewatch from 0.3.2 to 0.4.1

seanpoulter:
This pull request updates bundlewatch from 0.3.2 to 0.4.1.

### Test Plan

- The pipeline should pass
- I expect bundlewatch will add a check

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,7 +42,7 @@
         "babel-core": "7.0.0-bridge.0",
         "babel-loader": "8.3.0",
         "browserfs": "2.0.0",
-        "bundlewatch": "^0.3.2",
+        "bundlewatch": "^0.4.1",
         "cross-env": "6.0.0",
         "decompress": "^4.2.0",
         "diff-lines": "1.1.1",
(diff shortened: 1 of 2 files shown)


jcubic:
It looks like tests are failing because you run too many jobs in parallel

  'create tunnel' request failed with: URL https://saucelabs.com/rest/v1/***/tunnels - 400 (Bad Request) Error: HTTP request failed. Server response: 400: Too many active org tunnels for ***/_wmhilton: 5 >= 5 (ref: afda9700).

seanpoulter:
That's unfortunate. I made it run too quick and now we've got another problem. 😓

seanpoulter:
This version of `bundlewatch` is supposed to use Node.js v16 but there's a note it might work with earlier versions. That's great because we're still using Node.js v14. It works (despite the error):

https://github.com/isomorphic-git/isomorphic-git/actions/runs/14673957675/job/41186560841?pr=2079#step:6:138

image

<br>
and is reported in the checks:

image

seanpoulter:
> It looks like tests are failing because you run too many jobs in parallel
> 
> 'create tunnel' request failed with: URL https://saucelabs.com/rest/v1/***/tunnels - 400 (Bad Request) Error: HTTP request failed. Server response: 400: Too many active org tunnels for ***/_wmhilton: 5 >= 5 (ref: afda9700).

I reran those two since this is an unrelated change. It'll probably be worth confirming if open source can get more active connections, or if we've got to rate limit that job.

jcubic:
Should I merge this? I'm not sure why you don't have a way to merge. I was looking into permission, and you have write access same as for every repo in the org:

Przechwycenie obrazu ekranu_2025-04-26_15-43-40

As for node, if something doesn't work with old Node, we can remove support for v14 and v16.

I think that we don't need to support versions that reached the [end of life](https://endoflife.date/nodejs).

seanpoulter:
I'll try to limit the number of parallel BrowserStack test runs before you merge. I'm not sure if there's a way to limit it to up to 5 max for a job. 🤔 

As for Node.js versions that have reached end of life, yea, we can set that up and prepare for a major version bump.

seanpoulter:
The 3b51dae experiment worked as expected. 😆  If I understand correctly, GitHub does **not** have a mechanism to limit the number of parallel job runs globally to keep our BrowserStack test runs below the 5 parallel:
* `concurrency` can be used to ensure only one per group is running at a time.
* `strategy.max-parallel` limits it in the current PR/run, not globally

As a quick fix I'll set `max-parallel: 2` and hope we don't hit the limit. I'll check if the BrowserStack docs have a better solution.

seanpoulter:
This is working as expected. I'm happy to merge @jcubic.

As for the permissions, I have write access to the repo but I don't have access to push to the protected branch `main`. You'll probably have to add a bypass rule or grant me the "maintain" role. I found this table helpful https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/managing-repository-roles/repository-roles-for-an-organization#permissions-for-each-role that. You're looking for this row:

image
x

jcubic:
Added you a maintainer role. It's not very intuitive because it's in a completely different place than members.

seanpoulter:
> Added you a maintainer role. It's not very intuitive because it's in a completely different place than members.

That worked. I can merge. 🎉

seanpoulter:
I finally converged on the right environment variables so we don't download Chromium all the time. 🎉 

image

image

seanpoulter:
This is blocked by #2080.

seanpoulter:
This is ready for review @jcubic. 🙏 

Please let me know what you'd like to focus on next and if you'd like to coordinate off GitHub. I can send you an email to start.

jcubic (approved):
Looks good, you can merge. My email is public on my GitHub profile

PR #2080, opened 2025-04-27
chore: Improve BrowserStack test stability in the pipeline

seanpoulter:
This pull request:

- Limits the number of parallel BrowserStack tests to avoid hitting the account limit
- Fixes the conditional Chromium download. The download is skipped if the environment variables are set to anything (`null`, `''`, or `'false`).

### Test Plan

This had been tested and works:

image

image


jcubic (approved):

jcubic:
Looks good

Issue #1647, opened 2022-10-02
Git work with submodules that have been checked out by other means


danielr1996:
I'd like to propose a feature as well as a solution: 

Theoretically to reach 100% git compatibility this library should handle setup as well as working with checked out submodules. But for now lets focus on working with submodules that have been checked out by other means. 

As far as my understanding of submodules go to git dir is not stored as `.git`, instead `.git` is a file that points to the actual git directory. So is wrote this little workarround in Deno which works for me: 
```
let gitdir = '.git';
const stat =await Deno.stat(gitdir)
if (stat.isFile) {
    const dotgit = await Deno.readTextFile(gitdir)
    gitdir = path.join(Deno.cwd(), dotgit.replace('gitdir: ','').replace('\r','').replace('\n',''))
}

//@ts-ignore
const tagOid = await git.resolveRef({fs, dir: Deno.cwd(), gitdir, ref: 'main'});
```

Is this really all there is to support working with submodules? Or are there edge cases that I didn't consider. 
If it's really this simple I might create a PR if I *find the time™*

jcubic:
Unfortunately I have no idea, @mojavelinux maybe you know?

mojavelinux:
Yes, that's essentially it. If `.git` is a file rather than a directory, it points to a linked / embedded repository under the `.git` folder of the main repository. This linked repository can either be a linked worktree or a (sub)module. A linked worktree will contain the `commondir` and `gitdir` files, whereas a module will not.

Here's code I put together to test this logic:

```js
const fs = require('fs')
const { promises: fsp } = fs
const git = require('isomorphic-git')
const ospath = require('path')

;(async () => {
  let repo = { fs, dir: 'js' }
  const dotgit = ospath.join(repo.dir, '.git')
  const dotgitStat = await fsp.stat(dotgit).catch(() => ({ isFile: () => false, isDirectory: () => false }))
  if (dotgitStat.isFile()) repo = await resolveRepositoryFromSubmodule({ ...repo, gitdir: dotgit })
  console.log(await git.resolveRef({ ...repo, ref: 'HEAD' }))
})()

function resolveRepositoryFromSubmodule (repo) {
  return fsp
    .readFile(repo.gitdir, 'utf8')
    .then((contents) => contents.trimRight().substr(8))
    .then((submoduleGitdir) => {
      const gitdir = ospath.join(repo.dir, submoduleGitdir)
[...]

PR #2090, opened 2025-05-09
Support git commands run in submodules - draft

sdarwin:
Hi,

So far, this pull request adds submodule support to `currentBranch` and `resolveRef` as a first step towards including the same function in all the other command as well (https://isomorphic-git.org/docs/en/alphabetic)

Enable the ability to run git commands within submodule directories: fixes https://github.com/isomorphic-git/isomorphic-git/issues/1647

About worktrees:

It would definitely be nice to parse both submodules and worktrees. The complexity with worktrees is that sometimes git needs to know the worktree's own git directory (.git/worktrees/my_worktree) and other times the commondir ( ../.. ) . It varies. If each git function used one or other of those choices, it would be possible to go through all supported commands and set the selection to either 1 or 2 (worktree dir or commondir). However there is a third category - when information is needed from both. In those cases, it would require fiddling with the internals of each command or adding more variables inside the functions. Not impossible. However with limited availability of developer time there is value in just doing an easier step - supporting submodules - which would be immediately useful.  Nothing is blocking worktrees later, in the future.  

I have tested this and it works, but please review, let me know.  Then I could apply the code to other files.


jcubic:
You've broken all tests that use the function you modified.

sdarwin:
You are certainly right!
This code was based on the example in the issue, where it showed this snippet:  

```
const fs = require('fs')
const { promises: fsp } = fs
...  fsp.stat ...
```

Tests are failing on the `stat` command `Cannot read properties of undefined (reading 'stat')` .  

When trying this locally, I do have `require('fs')` and it succeeds, but that is not in the tests.

I was debating adding a warning earlier, and so should include it now. I am usually programming in python or bash, not nodejs.     But we need this feature, as end-users. The issue has been open multiple years, and it seems like fun to contribute.  

Your collaboration is more than welcome.   Do you have ideas?  I will try to investigate the errors.

jcubic:
`fs` inside the library is not the same as `require('fs');`. It's a wrapper over `fs` library for internal use.

If you want to contribute to a project, you first need to understand how it works, and if you modify something you need to test if it works. Not just the code snippet inside Node.js, but the library itself.

jcubic:
This is the code for the `fs` used internally:

https://github.com/isomorphic-git/isomorphic-git/blob/main/src/models/FileSystem.js

sdarwin:
Uploaded a new commit..

> Test if it works. Not just the code snippet inside Node.js, but the library itself.

Agreed.   Locally I have a copy of the library and then ran something like this when testing:  

```
const git = require('isomorphic-git')
let ref = git.resolveRef({
  fs: require('fs'),
  dir: '/opt/github/gitorg/gitrepo',
  fullname: true,
  ref: 'HEAD'
})
ref.then(function(result) {
     console.log(result) 
  })

```

src/api/resolveRef.js
```diff
@@ -34,10 +35,11 @@ export async function resolveRef({
     assertParameter('fs', fs)
     assertParameter('gitdir', gitdir)
     assertParameter('ref', ref)
+    const updatedgitdir = await discoverGitdir(fs, dir, gitdir)
 
     const oid = await GitRefManager.resolve({
       fs: new FileSystem(fs),
-      gitdir,
```
jcubic, line 40: `new FileSystem(fs)` is called twice for no reason.

jcubic (changes requested):
It would be nice to have some unit test that check if what you added actually work.

src/api/resolveRef.js
sdarwin: "new FileSystem(fs) twice"  
Should be fixed.

sdarwin:
> would be nice to have some unit test 

Yes.   
Not sure about my schedule though, have to see.

sdarwin:
To compose a "superproject" that has "submodules" it's necessary to check out multiple git repos.   
But they should all exist together in one directory.  
In the tests I am attempting to checkout the main project, and then a submodule, and copy the submodule in.
This works for a normal filesystem.  
It's failing on the browser filesystem.  
Any ideas?  
I'm wondering if BrowserFS were upgraded to ZenFS and/or Nodejs was upgrade to version 24, there would be improved "copy" commands available.

src/api/currentBranch.js
```diff
@@ -38,9 +39,11 @@ export async function currentBranch({
   try {
     assertParameter('fs', fs)
     assertParameter('gitdir', gitdir)
+    const fsp = new FileSystem(fs)
+    const updatedgitdir = await discoverGitdir(fsp, dir, gitdir)
```
jcubic, line 43: `discoverGitdir` is camel case, and `updatedgitdir` is all lowercase.

sdarwin:
Migrating everything from BrowserFS to ZenFS is likely somewhat complicated.    That is the latest and preferred filesystem so it should be done. Then 'symlink' would be available.  
In the meantime I have applied the submodules test to only the standard filesystem. Tests are passing.  CI should be passing as before.  
It's possible to logically examine the code and attempt to determine "is this correct?".   
Even if the browser filesystem is failing ( there is no reason to think that will fail ), adding submodule support in the cli filesystem is worthwhile.

jcubic:
Whatever you implemented needs to be tested that it's correct. It can be one test or multiple tests if there are multiple use case that your feature addresses.

jcubic:
Also, can you rebase with the latest changes?

jcubic:
But if you add your feature to 66 commands, then all should be tested.

sdarwin:
> The file system is abstracted away. If one is working, it will work for the other.

Ok, then it's sufficient to test on the regular filesystem, and we will assume the browser filesystem is functional, although it is not explicitly tested. This is reasonable. 

> clone recursively 

maybe not "clone recursively", we'll see, because that is getting into providing "full submodule support".  I am focusing on a subset of that, hopefully a stand-alone, self-sufficient subset of submodule features. Namely "support git commands run in submodules". Assuming you already have submodules checked out, you want to be able to run certain commands like "git branch".  Referring to the original Issue, the title is "work with submodules that have been checked out by other means". 

Therefore maybe this will be very limited, and won't affect 66 commands.  It will be the few commands that can be fixed by providing them the location of their git folder. Like `currentbranch` and `resolveref`.  I haven't gone through how many other commands fit into this category. What is the total number... Perhaps not so many. But it adds a usable feature and doesn't prevent other related features from being added later on.

sdarwin:
Removed the word "draft" from the description from the pull request. It's ready to test.

sdarwin:
Hi @jcubic, could you please review this pull request when you have a chance

sdarwin:
@jcubic ,  it appears the Github Actions didn't run on the latest commit.   

"2 workflows awaiting approval.  This workflow requires approval from a maintainer.   
[Learn more about approving workflows.](https://docs.github.com/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)"

Can you trigger another test run?

Also, by the way, test are expected to fail.  Compare to this unrelated one:  

https://github.com/isomorphic-git/isomorphic-git/actions/runs/16280340879

4 succeed and 4 failed.

jcubic:
Don't worry about the errors, only isomorphic-git-PR is important, this is an old pipeline from Azure.

I will need to review the code again.

sdarwin:
"support submodules".  Many end-users of our open-source software need this functionality.

Pinging @jcubic , the PR is ready to merge.

jcubic:
So we can update Node if want. But we should support versions that are maintained. Node 14 reached end of life. So if the feature you need is in Node 20 that is still maintained or 22 we can update Node to those versions. Node 24 may be problematic, since not everyone will use it, and we want tests for Node that people use.

We can also introduce ZenFS just for this one test, there was an idea to migrate (since BrowserFS is not maintained anymore), but it looks like too much effort for a single person (or maybe just no one want to do that).

I'm not sure what did you change in last commit because you squashed or amended the commit. It's easier to track changes if you don't modify the history. The commits are squashed on merge, so it doesn't matter what you have, but it's easier to track and maintain.

jcubic:
It looks like the test timeout because of the console.logs that slow down the tests.

jcubic:
I don't do much coding for this project, but I can see if I can migrate one of the tests to ZenFS. If it's straight forward, maybe we can use AI to migrate all the tests.

Your tests are problematic, I did not expect that the test will be just disabled for browser like this. I'm not sure what I expected, I probably just didn't think about it. Your feature is nice to have, but I would prefer to have tests that don't have any exceptions like this.

jcubic:
I didn't expect that you will run Linux shell commands as a string, and you will test like this.

I need to think what to do. I don't like the way tests are created, the condition `xit : it` and shell commands. But the tests are passing.

Maybe we should wait for the ZenFS migration. Over the weekend, I will try to see how hard it would be to migrate one test. Maybe it's all doable without much work.

I'm afraid that If I will merge, the tests will never be fixed. At least now, we have motivation to migrate to ZenFS.

jcubic:
The version of node is probably only in `package.json` and `azure-pipelines.yml`

jcubic:
The problem is that we don't know if the feature will work in the browser with ZenFS. Because we don't have tests that run in the browser.

I would prefer to wait for the ZenFS to be merged, and tests to be run in the browser.

jcubic:
As for PR with ZenFS, it's not ready to merge yet.

jcubic:
ZenFS got merged. You will need to merge with main and resolve conflicts. There were some minor changes in the code not related to tests. I should probably ask to extract them into a different PR.

jcubic (changes requested):
The code looks ok. I was expecting the AI to check the code, not tests.

One thing is to remove the shell exec, the browser test doesn't run at all (someone probably disable them.), but the `fs` module in the library is for a reason.

I don't recall when Browser/Device test were disabled, but I will check if I can turn them on again. There used to be Matrix run on all devices, now only one test runs.

PR #2114, opened 2025-06-17
fix: Handling of deleted trees on checkout

### LINKED ISSUE #2094: Checkout where a tree is removed can lead to deletion of ignored files
**Observed issue**

Assume you have branch `main`, and a branch `child` branched off from main at point A.
Then on `main`, you add and commit a folder with a `.gitignore` ignoring everything in the folder, except itself.
Any files created in this folder, is of course now ignored.

HOWEVER, checkout to the `child` branch now (which is at point A, and doesn't have the folder or gitignore), not only deletes the gitignore file, that is tracked only on `main` - but also *all* of the ignored files in the folder.

Standard git behaviour, is that these now appear *untracked* in the workdir.

**Solution**

Digging a little in the `commands/checkout.js`, I think this can be fixed by a couple of small changes (but haven't tested on any other cases, than the one above):

In the `analyze` function, the case in line 470 should return `rmdir-index` instead of `rmdir` (i.e. the tree was removed from the index, but it shouldn't automatically delete all files on disk - including any potentially ignored files):
```
        case '101': {
          switch (await stage.type()) {
            case 'tree': {
              // return ['rmdir', fullpath]   <-- current
              return ['rmdir-index', fullpath]
            }
```

And then I think in the `checkout` function, the handling of `rmdir` and `rmdir-index` is slightly wrong, and "flipped" re. what was intended - also when compared to the handling of `delete` and `delete-index`.

Currently it does:
```
        if (method === 'rmdir' || method === 'rmdir-index') {
          const filepath = `${dir}/${fullpath}`
          try {
            if (method === 'rmdir-index') {
              index.delete({ filepath: fullpath })
            }
            await fs.rmdir(filepath)
            if (onProgress) {
              await onProgress({
                phase: 'Updating workdir',
                loaded: ++count,
                total,
              })
            }
          }
```

Meaning it always deletes the directory, but only removes it from index, if it is explicitly a `remove-index`.
I would think it should be switched to:
```
        if (method === 'rmdir' || method === 'rmdir-index') {
          const filepath = `${dir}/${fullpath}`
          try {
            if (method === 'rmdir') {
              await fs.rmdir(filepath)
            }
            index.delete({ filepath: fullpath })
            if (onProgress) {
              await onProgress({
                phase: 'Updating workdir',
                loaded: ++count,
                total,
              })
            }
          }
```
I.e. always remove from index - but only if it is a *pure* `rmdir` should it also delete from disk. This also aligns with the behaviour for delete and delete-index:
```
        ops
          .filter(
            ([method]) => method === 'delete' || method === 'delete-index'
          )
          .map(async function([method, fullpath]) {
            const filepath = `${dir}/${fullpath}`
            if (method === 'delete') {
              await fs.rm(filepath)
            }
            index.delete({ filepath: fullpath })
            if (onProgress) {
              await onProgress({
                phase: 'Updating workdir',
                loaded: ++count,
                total,
              })
            }
          })
```

I can make a small PR for it, if you agree that is the intended behaviour - but full disclaimer, I haven't analyzed it thoroughly against the standard git behaviour in all cases. I have only looked at the specific case above!

gnillev:
## I'm fixing a bug or typo

Fixes #2094


jcubic (changes requested):
Can you write some unit tests for this? Something that will fail before the change?

gnillev:
> Can you write some unit tests for this? Something that will fail before the change?

I added a unit test.

In doing so, I also discovered, that it actually worked before in that case, but mainly relying on the fs implementation - specifically that `fs.rmdir` should throw ENOTEMPTY when it is not empty (thus skipping the deletion), which is caught and logged in `checkout`.
This helped me discover a discrepancy in our own FS implementation (not throwing ENOTEMPTY, and causing the deletion).

I still think the changes in this branch are more correct, when comparing to the behaviour of `delete` and `delete-index`, as described in the issue.

jcubic (approved):

jcubic:
If you find any other bug that you would like to fix, don't hesitate to create a PR. Remember about unit test, so it's easier to figure out what the PR is fixing.

Feature contribution are also welcome.

PR #2121, opened 2025-06-26
feat(merge --allow-unrelated-histories): adding parameter to merge unrelated branches

ARBhosale:
## I'm adding a parameter to an existing command merge:

- [x] add parameter to the function in `src/api/merge.js` (and `src/commands/merge.js` if necessary)
- [x] add a test case in `__tests__/test-merge.js` if possible

### Problem
Unable to `merge` when the two branches shared **no common ancestor**.  
In Git CLI this is handled with `git merge --allow-unrelated-histories`, but our API always threw `MergeNotSupportedError`, forcing users to resort to manual work-arounds.

### Solution
* **New flag** – `allowUnrelatedHistories` (default `false`) added to `merge`
* When no merge base is found **and** the flag is `true`, we fall back to the empty-tree OID (`4b825dc6…`) as a synthetic base so the standard three-way merge can proceed.
* **Docs updated** to describe the flag.
* **Tests added** covering:
  * Merging unrelated branches that add different files.
  * Merging unrelated branches that add the *same* filename with diverging content.

### Usage
```js
await git.merge({
  fs,
  dir,
  ours: 'main',
  theirs: 'other',
  allowUnrelatedHistories: true,
})

### DIFF
--- a/__tests__/test-merge.js
+++ b/__tests__/test-merge.js
@@ -1,6 +1,13 @@
 /* eslint-env node, browser, jasmine */
 const diff3Merge = require('diff3')
-const { Errors, merge, add, resolveRef, log } = require('isomorphic-git')
+const {
+  Errors,
+  merge,
+  add,
+  resolveRef,
+  log,
+  statusMatrix,
+} = require('isomorphic-git')
 const gitCommit = require('isomorphic-git').commit
 
 const { makeFixture } = require('./__helpers__/FixtureFS.js')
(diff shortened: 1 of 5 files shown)


jcubic (changes requested):
Can you test if both files are in repo (maybe with statusMatrix) and that both commits are there, by checking the messages or oids?

Also, this feature will require to have `--orphan` option when creating a branch. This is what git use in `checkout`.

jcubic (approved):
Looks good

PR #2144, opened 2025-08-05
perf: improve hashblob performance up to 100%

lukecotter:
## 50k Promises (Promise.all)

|            | **before (ms)** | **after (ms)** |
| ---------- | --------------- | -------------- |
| **Buffer** | 3822.918        | 1872.766       |
| **string** | 4072.27         | 2300           |

Buffer = 104.13% faster
string = 77.06% faster

### 50k Promises sequential

|            | **before (ms)** | **after (ms)** |
| ---------- | --------------- | -------------- |
| **Buffer** | 2821.764        | 1811.034       |
| **string** | 2509.682        | 1790           |

Buffer = 55.81 % faster
string = 40.21 % faster

- hashObject return a `Uint8Array`, no need to create another.
- hashBlob skip wrapping object in `Buffer` if it is already a `Uint8Array`
- `GitObject.wrap` : Create only 1 buffer instead of 3

### DIFF
--- a/src/api/hashBlob.js
+++ b/src/api/hashBlob.js
@@ -38,14 +38,14 @@ export async function hashBlob({ object }) {
 
     // Convert object to buffer
     if (typeof object === 'string') {
-      object = Buffer.from(object, 'utf8')
-    } else {
       object = Buffer.from(object)
+    } else if (!(object instanceof Uint8Array)) {
+      object = new Uint8Array(object)
     }
 
     const type = 'blob'
     const { oid, object: _object } = await hashObject({
-      type: 'blob',
+      type,
       format: 'content',
       object,
     })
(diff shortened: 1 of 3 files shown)


jcubic (approved):
Looks good

lukecotter:
@jcubic 

I was thinking about this and I think the change in hashBlob should probably have been this to be safer.

```

if (typeof object === 'string') {
      object = Buffer.from(object, 'utf8')
    } else if (!(object instanceof Uint8Array)) {
      object = Buffer.from(object)
    }
```

That is close to the original and will support more types but still skip wrapping for Buffer and Uint8Array. Given that the only types that should be provided are string and uint8array or subclasses it's probably not a big deal.

jcubic:
If you think this is better, you will need to create another PR.

Issue #2170, opened 2025-08-20
(WiP) PR: fix type and package bloat.


lemanschik:
At present the package produces individual files that contain more or less the same
even more bad we create none needed types for the cts files when we remove the default export 
of the src/index.js we get a named exports only compatible package for cjs and esm they can both share the same types so the index.d.cjs content would be like export * from './index';
so you see a lot more smart same in general for all packages 

you should almost always aim for named imports exports only to allow more easy mocking patching and consumption. 

let this also close the request for exposed API's in a type reduced way.
closes: #2032

PR #2171, opened 2025-08-20
Update index.js

lemanschik:
reduced type quirks and also module import require quirks now all get the same object.

## Features
- index.js now exports internal-apis.
- reduced redundant types via reexporting the bundled typed from the ESM build which are now valid for CJS and ESM.
- fixes PATH_NOT_EXPORTED errors
- adds support for extensionless imports via package.json default resolve extensionless with .js.

 
closes: #2170


lemanschik:
closing in favor of: npm install universal-git

PR #2212, opened 2025-10-04
fix(pull): Handle pull into empty repository

### LINKED ISSUE #1967: Git pull from empty repository.
Browser:
`Chrome Version 130.0.6723.93 (Official Build) (arm64)` 
with 
`@isomorphic-git/lightning-fs:4.6.0` 
`isomorphic-git:1.26.3`

In our situation, we're encountering an issue where the pull operation fails with the error message Could not find master., even though the remote repository does contain a master branch with commits.

```js
Could not find master.
    at GitRefManager.expand (index.js:2063:1)
    at async _merge (index.js:9568:1)
    at async _pull (index.js:9773:1)
    at async Object.pull (index.js:12129:1)
    at async gitPull (gitUtils.ts:156:1)
```

The issue appears to stem from the fact that our remote repository is set up as a bare repository. While it can be successfully cloned and pushed to, if this repository is updated through other ways, the first clone (at that moment origin was empty) will encounter failures during a pull operation. Similarly, performing a fetch operation (which completes without error) followed by a pull will result in the same failure.

Steps to reproduces:
1. have git bare repository
2. create clone A
3. create clone B
4. cloned B - commits/push to the origin successfully
5. cloned A - fails during doing `pull` or doing `fetch` first and then `pull` again

anish3333:
## Problem

When cloning an empty remote repository and then pulling, isomorphic-git fails with an error because there's no local branch to merge into. This is a common real-world scenario:

1. User initializes a Git repository on GitHub/GitLab
2. Clones it locally (empty clone - no commits, no branches)
3. Later, commits are pushed to the remote
4. User runs `git pull` to fetch those commits
5. **Current behavior:** Fails with error
6. **Expected behavior:** Creates local branch and successfully pulls (matches canonical Git)

## Solution

This PR adds empty repository detection and handling to `_pull()`:

### Key Changes

1. **Empty clone detection**: Check if `currentBranch()` fails/returns null before attempting merge
2. **Branch determination**: 
   - Use explicit `ref` parameter if provided
   - Fall back to remote's default branch (`refs/remotes/origin/HEAD`)
   - Fall back to `remoteRef` parameter
   - Throw clear error if none available (no hardcoded defaults)
3. **Branch creation**: Create local branch pointing to fetched commit
4. **Tracking setup**: Configure `branch.<name>.remote` and `branch.<name>.merge`
5. **Checkout**: Check out the newly created branch

### Canonical Git Compliance

This implementation strictly follows canonical Git behavior:
- No hardcoded preference for 'main' vs 'master'
- Respects remote server's advertised default branch
- Fails explicitly with `MissingParameterError` when branch cannot be determined
- Sets up identical tracking configuration

## Testing

Added 10 comprehensive tests covering:

**Core functionality:**
- ✅ Pull into empty clone with explicit `ref`
- ✅ Pull into empty clone using remote HEAD
- ✅ Pull into empty clone with `remoteRef` parameter
- ✅ Proper tracking configuration setup

**Error handling:**
- ✅ Error when branch cannot be determined (no guessing)
- ✅ Error when requesting non-existent branch

**Integration:**
- ✅ Subsequent pulls use tracking config (no `ref` needed)
- ✅ Branch list shows only created branch
- ✅ Working directory files checked out correctly
- ✅ Commit history integrity preserved
- ✅ Well-formed git config file
- ✅ Idempotency (multiple pulls don't break state)

All existing tests continue to pass.

## Documentation

Updated `pull.md` to document:
- Empty repository behavior section
- Clarified `ref` and `remoteRef` parameter descriptions for empty repo case
- Notes canonical Git compliance

## Example Usage

```javascript
// Initialize empty repository
await git.init({ fs, dir: '/repo' })
await git.addRemote({ 
  fs, 
  dir: '/repo', 
  remote: 'origin', 
  url: 'https://github.com/user/empty-repo.git' 
})

// Pull into empty repo (after remote has commits)
await git.pull({
  fs,
  http,
  dir: '/repo',
  ref: 'main',  // Creates local 'main' branch tracking origin/main
  author: { name: 'User', email: 'user@example.com' }
})

// Subsequent pulls work without ref parameter
await git.pull({ fs, http, dir: '/repo' })
```

## Breaking Changes

None. This is backward compatible:
- Existing pull operations continue to work unchanged
- Only adds new behavior for the previously failing empty clone scenario
- Error messages improved but error types remain consistent

## Related Issues

Fixes #1967

---

## Review Notes for Maintainers

**Key decision points to review:**

1. **Error behavior**: The implementation throws `MissingParameterError` when branch cannot be determined, rather than defaulting to 'main' or 'master'. This matches canonical Git's explicit failure behavior.

2. **Tracking config**: The implementation uses `GitConfigManager` to write branch tracking config. Please verify this is the preferred approach vs. using `setConfig()` API.

3. **Branch name normalization**: Helper function `normalizeBranchShort()` strips `refs/heads/` prefix. This handles both short names ('master') and full refs ('refs/heads/master').

4. **Remote HEAD reading**: The implementation attempts to read `refs/remotes/<remote>/HEAD` to determine default branch. This file is created by `git fetch` in canonical Git.

Please let me know if you'd like any changes to the approach or implementation details.

### DIFF
--- a/__tests__/test-pull.js
+++ b/__tests__/test-pull.js
@@ -148,270 +148,6 @@ describe('pull', () => {
     ])
   })
 
-  it('pull into empty clone', async () => {
-    // Setup - simulate cloning an empty repository
-    const { fs, gitdir, dir } = await makeFixture('test-pull-empty-clone')
-    await init({ fs, gitdir, dir })
-    await setConfig({
-      fs,
-      gitdir,
-      path: 'remote.origin.url',
-      value: `http://${localhost}:8888/test-pull-server.git`,
-    })
-    await setConfig({
-      fs,
-      gitdir,
-      path: 'remote.origin.fetch',
-      value: '+refs/heads/*:refs/remotes/origin/*',
-    })
-    // Delete HEAD to simulate empty clone with no branches
-    await fs.rm(path.join(gitdir, 'HEAD'))
-    // Test - verify no current branch exists (should throw)
-    let branch = null
-    try {
-      branch = await currentBranch({ fs, gitdir })
-    } catch (err) {
-      // Expected to fail when HEAD doesn't exist
-      expect(err.code).toBe(Errors.NotFoundError.code)
-    }
-    expect(branch).toBeNull()
-    // Pull should create the branch and fast-forward
-    await pull({
-      fs,
-      http,
-      gitdir,
-      dir,
-      remote: 'origin',
-      ref: 'master',
-      author: {
-        name: 'Mr. Test',
-        email: 'mrtest@example.com',
-        timestamp: 1262356920,
-        timezoneOffset: -0,
-      },
-    })
-    // Verify branch was created and contains commits
-    branch = await currentBranch({ fs, gitdir })
-    expect(branch).toBe('master')
-    const logs = await log({ fs, gitdir, dir, ref: 'refs/heads/master' })
-    expect(logs.map(({ commit }) => commit.message)).toEqual([
-      'Added c.txt\n',
-      'Added b.txt\n',
-      'Initial commit\n',
-    ])
-  })
-
-  it('pull into empty clone without ref parameter', async () => {
-    // Setup - simulate cloning an empty repository
-    const { fs, gitdir, dir } = await makeFixture(
-      'test-pull-empty-clone-no-ref'
-    )
-    await init({ fs, gitdir, dir })
-    await setConfig({
-      fs,
-      gitdir,
-      path: 'remote.origin.url',
-      value: `http://${localhost}:8888/test-pull-server.git`,
-    })
-    await setConfig({
-      fs,
-      gitdir,
-      path: 'remote.origin.fetch',
-      value: '+refs/heads/*:refs/remotes/origin/*',
-    })
-    // Delete HEAD and refs to simulate empty clone with no branches
-    await fs.rm(path.join(gitdir, 'HEAD'))
-    // Also need to ensure refs/heads/master doesn't exist
-    try {
-      await fs.rm(path.join(gitdir, 'refs', 'heads', 'master'))
-    } catch (e) {
-      // May not exist, that's fine
-    }
-    // Test - verify no current branch exists (should throw)
-    let branch = null
-    try {
-      branch = await currentBranch({ fs, gitdir })
-    } catch (err) {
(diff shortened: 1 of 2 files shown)


__tests__/test-pull.js
```diff
@@ -134,4 +147,809 @@ describe('pull', () => {
       'Initial commit\n ()',
     ])
   })
+
+  it('pull into empty clone', async () => {
+    // Setup - simulate cloning an empty repository
+    const { fs, gitdir, dir } = await makeFixture('test-pull-empty-clone')
```
jcubic, line 153: Where this fixture came from? I want to test it with canonical git. Becasue testing with existing repo after deleting .git/HEAD git throws error:

```
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

anish3333: Ok, I will test it with canoncial git repos as well and resolve the issues.

jcubic: Again, can you tell me where this fixture is located? All fixtures should be in `__tests__/__fixtures__` I don't see any directory like this.

fixtures is a a directory structure that are created for tests. You didn't add any fixture, only added fake name.

I'm not sure how it works, but all your tests should fail, because there are no fixtures like this.

anish3333: I have used the `makeFixture` function which basically creates a temporary repository as far as i understand

anish3333: @jcubic  I think I got it!

1. When `makeFixture('test-pull-empty-no-default')` runs, it doesn’t find a pre-existing fixture , basically for all the tests excluding the first three.
2. It creates an empty temp directory via `createTempDir()` used in makeNodeFixture()
3. The test calls init() and sets up the repo state manually

This is different from the first three tests (`test-pull`, `test-pull-no-ff`), which **require pre-made fixtures** with actual commit history.

**The pattern is basically:**

* Tests that need history → use fixture directories
* Tests that simulate empty clones → start with an empty temp repo, built dynamically in the test

So nothing is missing — those “fixtures” are just intentionally empty starting points.

jcubic:
In canonical git if you remove `.git/HEAD` or `.git/refs/heads/master` you have a broken git repository. You can't pull anymore.

PR #2302, opened 2026-04-22
Support windows file paths in utils/join.js

sdarwin:
Dan at Antora reported that discoverGitdir failed on Windows. That script depends on `join.js`.  
Update `join.js` to process Microsoft Windows file paths which use backslashes.
Please be sure all CI passes, although that is always done. In this case, I have not extensively tested the changes, observing the Azure results are important.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of Windows-style paths and drive prefixes (e.g., C:), preserving drive roots and reattaching drive letters after normalization.
  * Consistent separator normalization (backslashes → forward slashes) with preservation of trailing separators.
  * Path joining ignores empty segments and treats absolute arguments (including drive-letter absolutes) as resets, yielding Node-like absolute-path behavior.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,6 +80,7 @@ jobs:
 
       - name: Build artifacts
         if: ${{ !cancelled() && !github.event.pull_request.head.repo.fork }}
+        shell: bash
         run: |
           # Run the subset of the `test` necessary to test from the CLI.
           npx nps build.docs build.pack
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,7 +25,7 @@
         "isogit": "cli.cjs"
       },
       "devDependencies": {
-        "@isomorphic-git/cors-proxy": "^3.0.0",
+        "@isomorphic-git/cors-proxy": "^3.0.2",
         "@isomorphic-git/lightning-fs": "^3.3.0",
         "@isomorphic-git/pgp-plugin": "0.0.7",
         "@semantic-release/exec": "5.0.0",
@@ -877,7 +877,7 @@
       }
     },
     "node_modules/@isomorphic-git/cors-proxy": {
-      "version": "3.0.0",
+      "version": "3.0.2",
       "dev": true,
       "license": "MIT",
       "bin": {
--- a/package-scripts.cjs
+++ b/package-scripts.cjs
@@ -188,14 +188,8 @@ module.exports = {
       ),
       browsers: series.nps('test.chrome', 'test.firefox'),
       typecheck: 'tsc -p tsconfig.json',
-      setup:
-        process.platform === 'win32'
-          ? series.nps('gitserver.start')
-          : series.nps('proxy.start', 'gitserver.start'),
-      teardown:
-        process.platform === 'win32'
-          ? series.nps('gitserver.stop')
-          : series.nps('proxy.stop', 'gitserver.stop'),
+      setup: series.nps('proxy.start', 'gitserver.start'),
+      teardown: series.nps('proxy.stop', 'gitserver.stop'),
       node: process.env.CI
         ? (process.platform === 'win32'
             ? `cross-env ${jestEnv} ${retry3(jestCommand)}`
--- a/package.json
+++ b/package.json
@@ -118,7 +118,7 @@
     "simple-get": "^4.0.1"
   },
   "devDependencies": {
-    "@isomorphic-git/cors-proxy": "^3.0.0",
+    "@isomorphic-git/cors-proxy": "^3.0.2",
     "@isomorphic-git/lightning-fs": "^3.3.0",
     "@isomorphic-git/pgp-plugin": "0.0.7",
     "@semantic-release/exec": "5.0.0",
--- a/src/utils/join.js
+++ b/src/utils/join.js
@@ -2,6 +2,18 @@
  * This code for `path.join` is directly copied from @zenfs/core/path for bundle size improvements.
  * SPDX-License-Identifier: LGPL-3.0-or-later
  * Copyright (c) James Prevett and other ZenFS contributors.
+ *
+ * Windows support added:
+ *   - Backslashes are normalised to forward slashes before processing.
+ *   - Drive-letter prefixes (e.g. "C:") are detected and preserved through
+ *     normalisation, so absolute Windows paths are handled correctly.
+ *   - An absolute argument passed to join() resets the accumulated path,
+ *     matching Node behaviour and handling worktree gitdir paths properly.
+ *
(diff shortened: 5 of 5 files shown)


sdarwin:
regarding coderabbit's most recent comment,

"This one I'd push back on, for two reasons. First, the scenario is contrived. join('C:foo', '../') is a drive-relative path — a Windows concept meaning "relative to the current directory on drive C". Git never produces or consumes paths like that. Gitdir paths are either fully absolute (C:/path/to/.git) or plain relative (../something). Drive-relative paths simply don't appear in this codebase's inputs, so fixing their edge-case trailing-slash behaviour has no real-world impact here.
Second, the fix itself is debatable. Coderabbit's proposed change makes join('C:foo', '../') return 'C:' instead of 'C:/'. Neither is a meaningful or useful result for a path that has collapsed to nothing — you could argue 'C:/' is at least navigable, while 'C:' is ambiguous. It's not obviously more correct, just different.
My recommendation: skip it. Add a comment to the existing UNC/limitation note if you like, but don't change the code. The function is correct for every input pattern that actually occurs in isomorphic-git, and adding logic to handle a Windows edge case that git tooling never encounters just adds noise and something new to break. Coderabbit is doing its job, but not every technically-valid observation warrants a code change."

jcubic:
I think that the join function should work only for browsers. And a browser doesn't have drives on windows it works the same as on Linux and MacOs. The fix needs to be done in a different place. When using Node/Deno/Bun, it should use the built-in path.join instead of JavaScript code from zenfs.

sdarwin:
Those are two different options. Either the various client functions must do something more complicated and import multiple join functions, or adjust the current join function to be Windows compatible.
Claude preferred the second option, I think the second option is more elegant and universal, now it just works wherever it's needed.  Imagine you have any program, `grep`, `ls`, `pwd`, and with just a small change instead of supporting linux and macos, it also supports windows equally well, and then you can distribute it that way, it's very convenient, here the join function supports another OS such that surrounding programs using the utility can continue to use it, anyone who imports `join` knows it works.

jcubic:
I don't agree. And just because some AI said it's better, it's not an argument.

The built-in path.join already does what it needs to do on every platform it runs. Browser doesn't need Windows paths. That code was created for browsers only to optimize the bundle that is sent to the browser. There is no benefit of hardcoding Windows support to that code when you can just `import { join } from 'node:path'`;

sdarwin:
@jcubic wrote, a second time: "I don't agree."

I will close the PR for now.

jcubic:
If you want to fix the issue, you can do this. But I would prefer no code when the standard library does the job.

sdarwin:
> If you want to fix the issue, 

I don't understand your solution, and I don't agree with your solution.

Shouldn't there be a fix which affects all 86 files that use join.js?   That is, join.js should be adjusted, not discoverGitdir.   

Are you saying each individual client library separately?

If it is really as easy as:
import { join } from 'node:path';

That is one line, you as the maintainer could add this one line, and run the Azure tests.

jcubic:
This is not a one-liner because it needs to be loaded conditionally. This is just a reminder that the code to fix the issue you're fixing is already implemented as part of the standard library.

I just don't like patching the code that is supposed to run in the browser to fix a problem with the server.

sdarwin:
@jcubic Here is a proposed AI prompt, which I will run for a new pull request.  Tell me if you agree or not:

"""
Let's try a different tactic. The maintainer would prefer we call the official nodejs join function. How about this. We will go back to the earlier join.js, which I will upload again.  In join.js, very close to the top of the file, detect if the user is running inside a browser or using regular nodejs on a standard filesystem.  There are examples of this check other places in isomorphic-git. When inside a browser, continue with the same methodology as join.js is using now.  Nothing changes.   Or, when running regular normal nodejs, do something else. That is: "import { join } from 'node:path';" to import node's native join functionality from "path" and return that instead. (In the past I had difficulty importing from "path", not sure if package.json needs to be modified, or if this command works perfectly fine as-is. Maybe it's alright. Check if this step makes sense.)

With this change in place, all 80+ existing files that are doing this:

import { join } from '../utils/join.js'

will either be importing the status quo "browser" join function, or when they're running regular nodejs, not in a browser, it will be the normal nodejs path join function.  

Are the function signatures the same? Do you anticipate any problems? Most likely this solution will be fine.

"""

jcubic:
You don't need to create a prompt for this; And definilty don't need to modify every use of the join in the code base.

This is the code I had in mind:

```javascript
const isServer = typeof process !== 'undefined';

let nativeJoin;

if (isServer) {
  const { join } = await import("path");
  nativeJoin = join;
}

export function join(...args) {
  if (isServer) {
      return nativeJoin(...args);
  }
  if (args.length === 0) return '.'
  let joined
  for (let i = 0; i < args.length; ++i) {
    const arg = args[i]
    if (arg.length > 0) {
      if (joined === undefined) joined = arg
      else joined += '/' + arg
    }
  }
  if (joined === undefined) return '.'
  return normalize(joined)
}
```

sdarwin:
It's somewhat of a blunt hammer, but the idea is to compare the Windows results from this pull request against the `main` branch which currently shows:

```
Tests:       28 failed, 10 skipped, 1224 passed, 1262 total
```

sdarwin:
All the "Windows" tests should be green on this pull request.  And also the test labelled "Ubuntu".  Let's see. Please run github actions.

jcubic:
If you want to add `cors-proxy` it's better to include the `@isomorphic-git/cors-proxy` package. This will be a dev dependency. It doesn't make sense to copy/paste the code.

No new dependencies make sense for real dependencies.

sdarwin:
You are saying this should use "@isomorphic-git/cors-proxy".  I think cors-proxy is already a project dependency.  The problem is that cors-proxy is not Windows compatible.   I have submitted a PR to cors-proxy.  
- merge that other PR. or modify it
- release a new version of cors-proxy
- update isomorphic-git to use the new cors-proxy

Then we can continue with the PR here.

jcubic:
The Azure release of cors-proxy failed. I'm not sure what happened. It failed before you committed. It tried to create a tag that already exists.

sdarwin:
" It failed before you committed."
If the results are identical, before and after the commit, then it's at least some evidence the changes are acceptable.

jcubic:
Ok, I fixed the release; you can now install version 3.0.2 with your fix.

sdarwin:
Run actions

sdarwin:
In the last test run, just now, windows is/was 100% green. 
And this specific PR is focused on windows.

jcubic:
Sorry, didn't notice that the builds are green now.

PR #2304, opened 2026-04-24
Support windows file paths in utils/join.js - v2

sdarwin:
Support windows file paths in utils/join.js - v2
Check in CI

## Summary by CodeRabbit

* **Refactor**
  * Path-joining utility now uses the platform's native path join when running in server environments for more accurate server-side behavior.
  * The existing custom join + normalization is preserved for browser/bundler environments to maintain client-side compatibility.
  * Native implementation is loaded only on the server to avoid impacting client-side builds.

### DIFF
--- a/src/utils/join.js
+++ b/src/utils/join.js
@@ -4,6 +4,30 @@
  * Copyright (c) James Prevett and other ZenFS contributors.
  */
 
+// Detect whether we are running in a real Node.js environment (server-side),
+// as opposed to a browser or a browser bundler that polyfills `process`.
+// References: isomorphic-git uses this same pattern elsewhere (e.g. src/utils/pkg-info.js).
+const isServer =
+  typeof process !== 'undefined' && process.versions?.node != null
+
+// Load Node's native path.join when running in Node.js (CJS bundle).
+//
+// In the CJS Rollup output, `require` is always in scope. We guard with
+// `typeof require === 'function'` so that:
+//   - Rollup's static analyser sees a plain require('path') call and can mark
+//     'path' as an external CJS built-in (no bundling, no warning).
+//   - Browser bundles never reach this branch at runtime (isServer is false).
+//   - agadoo / tree-shake checks pass because there is no dynamic code eval.
+//
+// We intentionally use 'path' (not 'node:path') for maximum compatibility
+// with older Rollup versions and bundler configurations.
+let nativeJoin
+// eslint-disable-next-line no-undef
+if (isServer && typeof require === 'function') {
+  // eslint-disable-next-line no-undef
+  nativeJoin = require('path').join
+}
+
 function normalizeString(path, aar) {
   let res = ''
   let lastSegmentLength = 0
@@ -83,6 +107,9 @@ function normalize(path) {
 }
 
 export function join(...args) {
+  if (isServer && nativeJoin) {
+    return nativeJoin(...args)
+  }
   if (args.length === 0) return '.'
   let joined
   for (let i = 0; i < args.length; ++i) {


sdarwin:
@jcubic , here are a few problems with this strategy.

1. It's very difficult to import "path" to be able to use it.  The first attempt here failed, and then Claude is going through five attempts to solve that, and explaining why none of them will work.  Finally, wrapping the import in a function or something, multiple ways, which leads to point 2, next.  

2. tree shaking fails, because it doesn't like the fact the path import is opaque.

3. Critical issue: coderabbit says  "Native path.join produces backslashes on Windows — this regresses, not fixes, Windows support." All the implications of that statement should be thought through.

4. Coderabbit is suggesting:  

+function toPosix(p) {
+  return typeof p === 'string' ? p.replace(/\\/g, '/') : p
+}

which looks similar to the earlier PR, normalizing backslashes to forward slashes.
  
5.  The idea of "join" is not rocket science. It's just adding slashes in between segments of a path.  Why not have some confidence, and modify the current join.js the way it needs to be modified, and support that.  It would be different if the function was "render these images, parse these images to determine their contents" or something which is clearly extraordinarily difficult, and so you would need to outsource that, every time, and not maintain your own version of it.

PR 2302 was passing the tests, it standardizes on posix forward-slashes already, doesn't require a problematic import, it solves the issues.

jcubic:
PR 2302 was passing the tests because your code was not executed at all.

The current tests run on Linux/Node. Someone disabled the test matrix that was running on all supported platforms. I don't remember when this happened.

sdarwin:
> because your code was not executed at all

the code executed on Linux and passed which is more than "nothing at all" and is relevant.  Checking the historical CI situation, it seems microsoft windows CLI tests were not run before, so that is a separate issue from the test matrix. I am experimenting with adding windows tests. You might chat with Claude, writing verbose and extensive prompts, to debug the test matrix, that might be solved quickly.

sdarwin:
Now that GHA Windows tests exist I will rebase PR2302 and PR2304 from the main branch.  Re-run Github Actions.

There is a baseline to compare to, without any modification the results are:

```
Test Suites: 5 failed, 174 passed, 179 total
Tests:       28 failed, 10 skipped, 1224 passed, 1262 total
Snapshots:   190 passed, 190 total
Time:        187.905 s
Ran all test suites.
```

The interesting metric is "28 failed" on Windows in the status-quo. The current release of isomorphic-git shows 28 failures on Windows.

In the thread above, I commented "a few problems with this strategy", 5 items.  I think this method is too problematic to fix.   But we can see what's showing up with Windows tests.   I will push a commit in a moment.

sdarwin:
Oh, that's right.   It will not be able to even execute windows test. Because of these points:

1. It's very difficult to import "path" to be able to use it. The first attempt here failed, and then Claude is going through five attempts to solve that, and explaining why none of them will work. Finally, wrapping the import in a function or something, multiple ways, which leads to point 2, next.

2. tree shaking fails, because it doesn't like the fact the path import is opaque.

It's blocked at "tree shaking fails".

However, if it could get past tree-shaking I am confident the number of windows failures would increase rather than decrease.

jcubic:
If you will not be able to fix GitHub workflows for Windows, you can try to enable them in Azure DevOps. The only problem is that you probably will need to create a Microsoft account.

It may be less effort to make the Azure tests work with the matrix.

sdarwin:
There is functionally very little difference between running a script on Microsoft Windows from inside Github Actions  or on Azure.   
In either case, you are running on a Windows machine, and a script is executed.    
I explained the current error about tree-shaking above. It would not be solved by Azure.

jcubic:
Right now Azure Tests are failing anyway.

jcubic:
After your last changes, the build is failing. So the tests don't even run.

sdarwin:
"the build is failing".  I am aware of that. Discussed in more details in the above comments.  I believe the method in this PR won't work.

sdarwin:
Fixed in 2302.

PR #2306, opened 2026-04-26
checkout: surface silent rejections in batchAllSettled

andrewboldi:
## Problem

`batchAllSettled` in `src/commands/checkout.js` silently discards all `rejected` promise results from `Promise.allSettled`, returning only the fulfilled values. Combined with the function's outer `catch` that only `console.error`s and falls through to `return results`, **every individual file write that fails during checkout is swallowed without surfacing** — to either the caller or the user.

This contributes to a long-standing class of "clone appears to succeed but vault is empty / partially populated" bug reports across downstream consumers, especially on resource-constrained runtimes like iOS WKWebView where individual `fs.write` calls can fail under memory pressure mid-checkout.

Notable downstream reports referencing this exact symptom:
- Vinzent03/obsidian-git#642
- Vinzent03/obsidian-git#1012
- Vinzent03/obsidian-git#435
- Vinzent03/obsidian-git#475
- Vinzent03/obsidian-git#694

## Change

`batchAllSettled` now:
- Collects rejection reasons into a `rejections` array.
- `console.error`s each individual rejection with its stack (preserving previous diagnostic noise level for users running with verbose logging).
- After the batch loop, throws a single aggregated `Error` if any task rejected, including the count and first 5 error messages.
- The outer try/catch still logs the FATAL summary then **rethrows** instead of silently returning a partial `results` array.

This converts a class of silent failures into actionable errors. Behaviour for the all-success path is unchanged.

## Risk

Callers that were silently relying on partial-success results (i.e. swallowing a checkout that left files un-written) will now receive an exception instead of a misleading success. This is the intended fix — there should be no caller relying on the previous behaviour, since they had no way to know it was happening.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved batch operation error handling: per-task failures are now collected rather than suppressed, producing a summarized failure report (counts and concise messages, including stack info when available). All task rejections are logged with clearer context, and fatal batch errors surface as an aggregated error for easier troubleshooting.

### DIFF
--- a/src/commands/checkout.js
+++ b/src/commands/checkout.js
@@ -8,6 +8,7 @@ import { _walk } from '../commands/walk.js'
 import { CheckoutConflictError } from '../errors/CheckoutConflictError.js'
 import { CommitNotFetchedError } from '../errors/CommitNotFetchedError.js'
 import { InternalError } from '../errors/InternalError.js'
+import { MultipleGitError } from '../errors/MultipleGitError.js'
 import { NotFoundError } from '../errors/NotFoundError.js'
 import { GitConfigManager } from '../managers/GitConfigManager.js'
 import { GitIndexManager } from '../managers/GitIndexManager.js'
@@ -279,7 +280,7 @@ export async function _checkout({
       await GitIndexManager.acquire(
         { fs, gitdir, cache, allowUnmerged: true },
         async function (index) {
-          await Promise.all(
+          const settled = await Promise.allSettled(
             ops
               .filter(
                 ([method]) =>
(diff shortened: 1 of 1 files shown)


jcubic:
Thanks for the fix. The code looks good. Can you apply changes suggested by CodeRabbit?

If you don't agree with the suggestions, please give a reason.

andrewboldi:
@jcubic thanks for the review! Pushed 40d73a38 addressing CodeRabbit's feedback:

1. **Major** — extended the rejection collection to the default (blocking) checkout branch (lines 278-354). The original `try { ... } catch (e) { console.log(e) }` per-file silent swallow is now replaced with `Promise.allSettled` + aggregated `MultipleGitError`, so write failures now actually fail the checkout for default callers (the failure mode this PR was originally meant to surface).

2. **Inline (749-757)** — `batchAllSettled` now throws `MultipleGitError` instead of a plain `Error`, so callers can introspect `.errors` / `.data` programmatically (matches the pattern in `src/api/add.js`).

3. **Nitpick (760-764)** — removed the outer `try/catch` that double-logged the synthetic aggregate; per-task rejection logs in the loop are sufficient.

4. **Inline (732-736)** — replaced inline ternary with `result.reason?.stack ?? result.reason` to satisfy Prettier.

ESLint and Prettier both pass locally on `src/commands/checkout.js`.

Issue #2301, opened 2026-04-17
v1.37.5 Breaks Browser usage when using Vite


amxmln:
I am using `isomorphic-git` in the browser with Vite. Since the `v1.37.5` release, cloning fails because the release introduced a dependency on the `crypto`-module which is Node only. Vite does not polyfill these modules, see here for more information: [Module externalized for Browser Compatability](https://vite.dev/guide/troubleshooting.html#others)

There might be a way to manually polyfill the crypto module for browsers, but then I feel like the version should have been a major one, since it introduces breaking changes.

amxmln:
Some updates on my part here, using `vite-plugin-node-polyfills` to polyfill the `crypto`, `stream` and `util` packages I was able to get cloning to work in the browser again.

However, these were not necessary before, which makes me think that #2299 perhaps somehow caused this library to use the Node.js builds even in browsers. Could that be the case?

The PR explicitly mentions that the improvements should only affect Node.js-builds.

I don’t think having to include these polyfills should be necessary for browser builds, especially since during the inspection of the built files I couldn’t actually find anything suggesting they’re necessary if the browser build is used. They nearly double my bundle size.

jcubic:
The library imports the `crytpo` library, so it imports the Node module. We need to implement (copy) **sha1** and embed it in the code.

Do you want to help with this? Those polyfills should not be needed. I'm not sure about the `stream` and `util`.

The problem is that someone did remove the automated tests that were running for all platforms, including browsers and phones. And now tests only run in Node. I'm not sure how to enable them back. Maybe we should ask an AI agent to find this in Git history.

amxmln:
Some more updates:

It looks like Vite tries to load `node_modules/isomorphic-git/index.cjs` when bundling, which does use Node APIs (see Line 11: `var crypto$1 = require('crypto');`).

Changing `package.json` to:

```json
"exports": {
    ".": {
      "types": "./index.d.ts",
      "default": "./index.js"
    },
   ...
}
```

fixes the issue (it only imports the browser build, which doesn’t contain the `crypto` import.

So I guess technically everything is working correctly, but the module resolution has been borked (in the sense that the Node build was always loaded) and nobody noticed before?

I was thinking that fixing the resolution somehow to import `index.js` for the browser and `index.cjs` for node might work, but I am not very familiar with how that works.

Something like:

```json
"exports": {
    ".": {
      "import": {
        "types": "./index.d.ts",
        "default": "./index.js"
[...]

PR #2309, opened 2026-05-08
Fix browser entrypoint not being used in some non-node build contexts

amxmln:
As discussed in #2301, this PR sets the base import to only import the `.cjs` versions in `node` contexts and  defaults to the `.js` versions of the code otherwise.

The `.js` versions don’t include any Node-only APIs and thus run in browsers as intended.

Like I mentioned before, it’s a bit of a shot in the dark, but it seems to work on my setup and I’m hoping the existing tests can confirm that it doesn’t break anywhere else.

## I'm fixing a bug or typo

## Summary by CodeRabbit

* **Chores**
  * Added new contributor to acknowledgments
  * Updated package export configuration for improved compatibility

* **Documentation**
  * Updated contributors list

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -998,6 +998,16 @@
       "contributions": [
         "security"
       ]
+    },
+    {
+      "login": "amxmln",
+      "name": "Amadeus Maximilian",
+      "avatar_url": "https://avatars.githubusercontent.com/u/15271679?v=4",
+      "profile": "https://github.com/amxmln",
+      "contributions": [
+        "code",
+        "bug"
+      ]
     }
   ],
   "commitConvention": "angular"
--- a/README.md
+++ b/README.md
@@ -418,6 +418,7 @@ Thanks goes to these wonderful people ([emoji key](https://github.com/kentcdodds
   </tr>
   <tr>
     <td align="center"><a href="https://github.com/N0zoM1z0"><img src="https://avatars.githubusercontent.com/u/161784452?v=4?s=60" width="60px;" alt=""/><br /><sub><b>N0zoM1z0</b></sub></a><br /><a href="#security-N0zoM1z0" title="Security">🛡️</a></td>
+    <td align="center"><a href="https://github.com/amxmln"><img src="https://avatars.githubusercontent.com/u/15271679?v=4?s=60" width="60px;" alt=""/><br /><sub><b>Amadeus Maximilian</b></sub></a><br /><a href="https://github.com/isomorphic-git/isomorphic-git/commits?author=amxmln" title="Code">💻</a> <a href="https://github.com/isomorphic-git/isomorphic-git/issues?q=author%3Aamxmln" title="Bug reports">🐛</a></td>
   </tr>
 </table>
 
--- a/package.json
+++ b/package.json
@@ -8,8 +8,14 @@
   "module": "./index.js",
   "exports": {
     ".": {
-      "types": "./index.d.cts",
-      "default": "./index.cjs"
+      "node": {
+        "types": "./index.d.cts",
+        "default": "./index.cjs"
+      },
+      "default": {
+        "types": "./index.d.ts",
+        "default": "./index.js"
+      }
     },
     "./http/node": {
       "import": {


jcubic:
How can I test this locally? Can you prepare a code that I can run to validate that this fails before the fix and it's working after?

amxmln:
Sure, here’s a small example illustrating the issue and to test the fix once applied: https://github.com/amxmln/isomorphic-git-vite-tester

jcubic:
Thanks for the repo. You need to give me some time. I need to check different ways the library is imported. It needs to work as before. I don't think that this is tested in any way in CI/CD. I don't want to break any existing code.

If you want to help (and speed things up), you can try to create a repo for different ways the library is used from the README and the examples in the docs.

You can use a single Git repo and put different types into each directory. This would help a lot. Thanks.

amxmln:
Sure, I’m happy to help out with this.

Here’s the different cases I’d prepare:

- [X] Node with `require`
- [X] Node with `import`
- [X] Bundling with Vite
- [X] Web Workers with Vite
- [X] Bundling with Webpack (?)

Are there other cases I might have missed or that you’d like to see? 🤔 

**EDIT:** I have updated the linked repo with examples for the checked off cases. So far, I couldn’t detect any issues.

jcubic:
Also using jsDelivr.com and Unpkg.com URLs. They should work with just the path to the package without the need to select the file.

amxmln:
I’ve added two examples for unpkg and jsDelivr to the repo. However, I am not sure how to point the CDNs to my changed PR, do you have any pointers towards that?

Reading the details from [Unpkg](https://unpkg.com/), it seems it will pick up the `"default"` property of the export, so it should still keep working.

As for jsDelivr, something like `<script src="https://cdn.jsdelivr.net/npm/isomorphic-git"></script>` doesn’t seem to work (opening it in a browser downloads an unminified index.cjs, I think matching the one defined in the `main`-field or the current `exports['.'].default`-field of `package.json`).  Their [documentation](https://www.jsdelivr.com/documentation#id-configuring-a-default-file-in-packagejson) seems to suggest it would be the `main`-field.

To me this suggests that using jsDeliver without the need to select a file doesn’t work with any current version of `isomorphic-git` (or `lightning-fs` for that matter), or am I missing something?

jcubic:
JSDevlir can point to GitHub and your fork:

```
https://cdn.jsdelivr.net/gh/<user>/<repo>@<branch>
```

But I'm not sure if it works the same as for NPM and if you need to use the path to the file.

amxmln:
Thanks for the hint and sorry for the wait. I tried using the URL you suggested and it complains about not being able to determine the version. I guess I’d have to somehow cut a release in my fork or something for it to work? Because the [CDN-URL](https://cdn.jsdelivr.net/gh/amxmln/isomorphic-git@fix%2Fbrowser-entrypoint/) just points to the repo itself right now and not any built files.

I do have a few more findings after adding a Webpack-Example:

- Webpack v5 fails with the current `package.json`, but will work once my fix is applied
- The `package.json` contains an `unpkg`-field which already points to the correct (browser-based) file path regardless of my fix

jcubic:
Yeah, probably the CDN that points to the Git repo was not a good idea. The build files are located in the tar file. You will need to have a repo with the content of the tar file.

I will check the repo and confirm if it works. I think that it will be ok to just test after we release a new version.

Every PR triggers a release, so we will need to create a few releases to test this. I'm not sure if there is another way.

jcubic:
Checked locally, all fine. After merge and release comment. We will check the CND files.

amxmln:
Thanks for merging! I just re-ran the unpkg and jsDelivr examples and they both succeed. However, I am still not able to get the jsDelivr example to run without adding a path to `index.umd.min.js`.

Maybe to make that work, we’d need to add a `jsdelivr` key to `package.json` like we do for unpkg?

jcubic:
It seems that they support it, this is in documentation: https://www.jsdelivr.com/documentation#id-configuring-a-default-file-in-packagejson

## Configuring a default file in package.json
For packages hosted on npm, we support serving "default" files with shorter URLs. The default file can be configured by setting one of the following fields in package.json, with jsdelivr having the highest priority:

* jsdelivr
* browser
* main

We will first attempt to locate a minified version of the file provided here (by removing the extension and looking for the same file .min.js). If we can't find one, we will minify ourselves.

Be advised that you must include the file extension in the values, for example:

```
"main": "./index" // this will NOT work
"main": "./index.js" // this is the correct way
```

For projects having both a JS and a CSS file, use one the above fields for JS and a style field for the CSS file ([example](https://cdn.jsdelivr.net/npm/bootstrap@5.3.2/package.json)).

## Restrictions
* Packages larger than 150 MB or single files larger than 20 MB (in the case of GitHub) are not supported by default. We recommend removing files that are not needed from your package when possible. If you need to set a higher limit for your package, open an issue in this repo.
* HTML files are served with Content-Type: text/plain for security reasons.

You can create anotehr PR.

PR #2345, opened 2026-05-26
Improve internal error reporting guidance

### LINKED ISSUE #2216: Improve fatal error message
By default, the fatal error message just suggest creating an issue in this project.

This is not ideal because people report bugs in applications they are using. Without any information what kind of application it is.

Those bugs can be reported here, but also in the project the user is using. The developers of the app that use isomorphic-git should fix the bug. And if they found that the issue is on our side, they should report a bug themselves with all required info.

So the text is to improve the error message somehow. To make it clear that if they are not developers and use some application, they should report the bug to the developers of that app.

Example error and bug report:
* https://github.com/isomorphic-git/isomorphic-git/issues/2140

Proposed information that should be included:
> "If you're not a developer, report the bug to the developers of the application you're using. If this is a bug in isomorphic-git they should create a proper bug themselves. The bug should include a minimal reproduction and details about the version and environment."

Will-thom:
## Summary
- clarify InternalError guidance for application users vs developers
- keep isomorphic-git issue reporting guidance targeted to developers with reproduction details
- add error-message coverage in GitError tests

Closes #2216

## Tests
- npx nps lint
- npx prettier --check src/errors/InternalError.js __tests__/test-GitError.js __tests__/test-GitError-in-submodule.js
- $env:NODE_OPTIONS='--experimental-vm-modules --max-old-space-size-percentage=80'; npx jest __tests__/test-GitError.js __tests__/test-GitError-in-submodule.js --runInBand --ci --coverage=false
- npx nps test.typecheck

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messages with clearer, actionable guidance for users regarding issue reporting and for developers regarding required information for bug reports, including version, environment details, and reproduction steps.

* **Tests**
  * Added test cases to verify error message formatting and ensure proper guidance text is included for both end-users and developers.

[Review Change Stack](https://app.coderabbit.ai/change-stack/isomorphic-git/isomorphic-git/pull/2345?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/__tests__/test-GitError-in-submodule.js
+++ b/__tests__/test-GitError-in-submodule.js
@@ -33,4 +33,15 @@ describe('Errors', () => {
       }
     `)
   })
+  it('create an InternalError with actionable reporting guidance', async () => {
+    const e = new Errors.InternalError('Something unexpected happened.')
+
+    expect(e.message).toContain(
+      "If you're using an application that depends on isomorphic-git"
+    )
+    expect(e.message).toContain(
+      "If you're a developer and you believe this is a bug in isomorphic-git"
+    )
+    expect(e.message).toContain('Something unexpected happened.')
+  })
 })
--- a/__tests__/test-GitError.js
+++ b/__tests__/test-GitError.js
@@ -33,4 +33,15 @@ describe('Errors', () => {
       }
     `)
   })
+  it('create an InternalError with actionable reporting guidance', async () => {
+    const e = new Errors.InternalError('Something unexpected happened.')
+
+    expect(e.message).toContain(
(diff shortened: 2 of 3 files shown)


jcubic:
Thanks for improving the error message.

jcubic:
The fix to that issue was actually already applied. The error messages already mention the application that use isomorphic-git.

The old message (that was in the issue) only had CTA to report a bug in isomorphic-git. Because of this, when one app had a bug, we had a flood of bug reports.

But it's always good to mention reproduction.

Issue #1867, opened 2024-03-04
How can I abort a push operation on Node? AbortController seems to be not supported.


ipeychev:
Hey guys, I have a question, please: how can I abort a push operation on Node?
Context: there is too much data sometimes, and the operation may take too much time. I want to give the user the opportunity to abort the operation. I though to use AbortController, but it seems it is not supported?

Best regards,

jcubic:
There is no way to abort any git operation with isomorphic-git AFAIK.

If you want to add this feature, you will need to look into code. If you add this I'm happy to merge the PR.

ipeychev:
hm, okay. I can try to do it for push `command`, and we can see if that will be feasible actually.

ipeychev:
I sent a PR: https://github.com/isomorphic-git/isomorphic-git/pull/1880
I needed to downgrade some packages, but now the tests are passing on my Mac and on Azure/Linux. Of course, there were some broken tests before, they are still failing, but this is another story.

PR #2369, opened 2026-06-30
feat(http): support custom fetchOptions in built-in HTTP clients

### LINKED ISSUE #1246: Support custom Fetch options in http/web
I suppose this also applies to `http/node` but I specifically ran into this limitation with `http/web`. It's possible to customize request headers in functions like [`clone`](https://isomorphic-git.org/docs/en/clone). However, it is _not_ possible to supply custom [fetch](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch) options since the current `http` client [has no such argument](https://github.com/isomorphic-git/isomorphic-git/blob/main/src/http/web/index.js#L23).

One use case would be using an internal `corsProxy` that has its own non-token auth such as session/cookie or Basic Auth. To authenticate against the API, `fetch` would need to have `credentials: 'include'` option. Without it, the browser wouldn't forward the HTTP-only session cookie or the Basic Auth credentials. (The `headers` param described [here](https://isomorphic-git.org/docs/en/onAuth#option-2-headers) only works for token-based auth.) 

As a workaround, I "forked" [http/web/index.js](https://github.com/isomorphic-git/isomorphic-git/blob/main/src/http/web/index.js) into our project. However, this seems inefficient since `credentials: 'include'` is the only substantive change I had to make. Furthermore, I had to copy-paste a few util functions which, if this workaround is encouraged, should ideally be exported from the lib in [src/index.js](https://github.com/isomorphic-git/isomorphic-git/blob/main/src/index.js).

Could we make the built-in HTTP client more robust? Here are the [`fetch` options](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch#Syntax) once again for reference.

JessenReinhart:
## Closes #1246 #1867

The built-in HTTP clients (`http/web`, `http/node`) hard-coded the options passed to the underlying transport, leaving no escape hatch for callers that need extra knobs (e.g. `credentials: 'include'` for a CORS proxy, `signal` for AbortController, `timeout` for simple-get, etc.).

## Change

Add an optional `fetchOptions` field to `GitHttpRequest`. When provided, the object is spread into the underlying transport call. When omitted, behaviour is unchanged (backward compatible).

### `src/http/web/index.js`
```js
const res = await fetch(url, { method, headers, body, ...fetchOptions })
```

### `src/http/node/index.js`
```js
get({ url, method, headers, agent, body, ...fetchOptions }, ...)
```

## Files changed
- `src/http/web/index.js` — accept and spread `fetchOptions`
- `src/http/node/index.js` — accept and spread `fetchOptions`
- `src/typedefs-http.js` — document `fetchOptions` in `GitHttpRequest`
- `__tests__/test-httpClient.js` — new test file (web: mock global fetch; node: integration smoke)

## Backward compatibility

The default `fetchOptions = {}` makes this a no-op for all existing callers.

## Example use case

```js
await git.clone({
  ...,
  http: {
    ...git.http,
    request: ({ url, ...rest }) => git.http.request({
      url, ...rest,
      fetchOptions: { credentials: 'include' }
    })
  }
})
```

## Summary by CodeRabbit

* **New Features**
  * Added an optional `fetchOptions` parameter to HTTP requests, allowing callers to pass extra request configuration in both web (`fetch`) and Node (`simple-get`) environments.
* **Tests**
  * Expanded automated coverage to confirm `fetchOptions` are correctly forwarded, and that existing behavior remains unchanged when `fetchOptions` is not provided.
* **Documentation**
  * Updated the HTTP request type definition to include the new optional `fetchOptions` field.

### DIFF
--- a/__tests__/test-httpClient.js
+++ b/__tests__/test-httpClient.js
@@ -68,7 +68,10 @@ describe('httpClient', () => {
       try {
         const { request: webRequest } = await import('../src/http/web/index.js')
         await webRequest({
+          url: 'https://example.com',
+          method: 'GET',
           headers: {},
+          fetchOptions: {}, // explicitly empty for backward compatibility
         })
       } finally {
         globalThis.fetch = originalFetch
--- a/__tests__/test-httpClient.js
+++ b/__tests__/test-httpClient.js
@@ -1,14 +1,14 @@
 /* eslint-env node, browser, jasmine */
 import { jest } from '@jest/globals'
 
+import { request as nodeRequest } from '../src/http/node/index.js'
 import { request as webRequest } from '../src/http/web/index.js'
 
-const simpleGetMock = jest.fn()
-await jest.unstable_mockModule('simple-get', () => ({
+const mockSimpleGet = jest.fn()
+jest.mock('simple-get', () => ({
   __esModule: true,
-  default: simpleGetMock,
+  default: mockSimpleGet,
 }))
-const { request: nodeRequest } = await import('../src/http/node/index.js')
 
 describe('httpClient', () => {
   describe('web', () => {
@@ -89,7 +89,7 @@ describe('httpClient', () => {
 
   describe('node', () => {
     it('passes fetchOptions through to simple-get', async () => {
-      simpleGetMock.mockImplementation(
+      mockSimpleGet.mockImplementation(
         /**
          * @param {any} opts
          * @param {any} cb
@@ -110,7 +110,7 @@ describe('httpClient', () => {
         headers: {},
         fetchOptions: { timeout: 5000, family: 4 },
       })
-      expect(simpleGetMock).toHaveBeenCalledWith(
+      expect(mockSimpleGet).toHaveBeenCalledWith(
         expect.objectContaining({
           timeout: 5000,
           family: 4,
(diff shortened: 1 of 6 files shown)


jcubic:
Typechecking errors in CI/CD:

```
nps is executing `test.typecheck` : tsc -p tsconfig.json
src/http/node/index.js(25,5): error TS2739: Type 'Buffer<ArrayBuffer>' is missing the following properties from type 'AsyncIterableIterator<Uint8Array<ArrayBufferLike>>': [Symbol.asyncIterator], next
src/http/web/index.js(22,5): error TS2739: Type 'Uint8Array<ArrayBuffer>' is missing the following properties from type 'AsyncIterableIterator<Uint8Array<ArrayBufferLike>>': [Symbol.asyncIterator], next
src/http/web/index.js(26,17): error TS2774: This condition will always return true since this function is always defined. Did you mean to call it instead?
src/http/web/index.js(31,42): error TS2339: Property 'entries' does not exist on type 'Headers'.
src/http/web/index.js(36,17): error TS2339: Property 'method' does not exist on type 'Response'.
src/utils/asyncIteratorToStream.js(4,11): error TS2305: Module '"readable-stream"' has no exported member 'PassThrough'.
```

jcubic:
Let me know when it's ready to merge.

That eslint ignore is weird. It's not related to the change you've made. And the tests on main are working fine.

PR #2382, opened 2026-07-20
fix(statusMatrix): do not traverse symlinks in GitWalkerFs (#1215)

### LINKED ISSUE #1215: statusMatrix() does not work correctly when there is a symlink that targets an ignored directory
## Bug description

When there is a symlink that targets an ignored directory (such as `node_modules`), `statusMatrix()` would list files within the ignored directory as "new". This is different from `git status --porcelain=v1` when these files are not listed.

## Steps to reproduce

1. Clone https://github.com/KSXGitHub/isomorphic-git-issue-incorrect-file-listing-for-symlink.git
2. Run `npm install` and `npm test`

You will see that the files "within" `symlink-to-ignored-directory` are listed by isomorphic-git but not by native git.

minutechreview:
## Description
Fixes #1215.

When a repository contains a symlink targeting an ignored directory (such as `node_modules`), `statusMatrix()` was previously traversing the symlink target because `GitWalkerFs.readdir()` called `fs.readdir()`, which follows directory symlinks. As a result, files within the ignored directory were incorrectly reported as untracked/new files.

This change adds a check in `GitWalkerFs.readdir()` to ensure `readdir` is only performed on `tree` entries (actual directories), matching the behavior of `GitWalkerRepo` and `GitWalkerIndex`.

## Changes Made
1. **`src/models/GitWalkerFs.js`**: Added `if ((await entry.type()) !== 'tree') return null` inside `GitWalkerFs.readdir(entry)`.
2. **`__tests__/test-statusMatrix.js`**: Added unit test `statusMatrix with symlink targeting ignored directory` to prevent regression.

## Verification
- Added unit test passes cleanly in Jest.
- Tested against reproduction case, confirming `statusMatrix()` output matches native `git status --porcelain=v1`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved directory reading behavior by safely short-circuiting when encountering non-directory entries.

* **Tests**
  * Added coverage for status matrix behavior when a symlink points to an ignored directory, ensuring expected paths and status/stage counts are produced.

### DIFF
--- a/__tests__/server-only.test-statusMatrix-symlink.js
+++ b/__tests__/server-only.test-statusMatrix-symlink.js
@@ -1,0 +1,22 @@
+/* eslint-env node, jasmine */
+import * as path from 'path'
+
+import { statusMatrix } from 'isomorphic-git'
+
+import { makeFixture } from './__helpers__/FixtureFS.js'
+
+describe('statusMatrix with symlink', () => {
+  it('statusMatrix with symlink targeting ignored directory', async () => {
+    const { fs, dir, gitdir, _fs } = await makeFixture('test-empty')
+    await fs.write(path.join(dir, '.gitignore'), 'ignored-dir\n')
+    await fs.mkdir(path.join(dir, 'ignored-dir'))
+    await fs.write(path.join(dir, 'ignored-dir', 'file.txt'), 'secret')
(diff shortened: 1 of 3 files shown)


jcubic:
Please run:

```
npx eslint --fix __tests__/server-only.test-statusMatrix-symlink.js
```

to fix linting errors

jcubic:
Thanks for the PR.

PR #2385, opened 2026-07-22
feat(log): include changed file object ids

### LINKED ISSUE #1919: Add filesChanged along with the no of lines insertions/deletions (--numstat) to `git.log`
It would be really useful to include the list of files changed (+/- insertions/deletions) for each commit in `git.log`.
I'm building a git viz tool which currently relies on server side git for performing this operation

minutechreview:
Fixes #1919.\n\nAdds the maintainer-proposed `includeChanges` option to `git.log()`. When enabled, each returned `commit` includes `changes` tuples in the form `[newOid, oldOid, filepath]`, comparing the commit against its first parent (or the empty tree for an initial commit). Added and removed files use `null` for the missing oid.\n\nThis intentionally exposes object ids only; callers can pass the corresponding blob contents to any diff library they choose.\n\nTests:\n- `npx nps lint`\n- `node --experimental-vm-modules node_modules/jest/bin/jest.js __tests__/test-log.js`\n- declaration build and TypeScript typecheck

## Summary by CodeRabbit

* **New Features**
  * Added an `includeChanges` option to commit history queries.
  * When enabled, each returned commit includes `commit.changes` as `[newOid, oldOid, filepath]` tuples for added, modified, and removed paths.
  * Supports edge cases like replacing a file with a directory and correctly reports changes at shallow-history boundaries.
* **Documentation**
  * Updated the commit object schema to describe the optional `changes` field and its tuple format.
* **Tests**
  * Expanded end-to-end, fixture-driven coverage for `includeChanges`, including shallow boundary behavior.

### DIFF
--- a/src/commands/log.js
+++ b/src/commands/log.js
@@ -1,7 +1,9 @@
 // @ts-check
 import '../typedefs.js'
 
+import { TREE } from '../commands/TREE.js'
 import { _readCommit } from '../commands/readCommit.js'
+import { _walk } from '../commands/walk.js'
 import { NotFoundError } from '../errors/NotFoundError.js'
 import { GitRefManager } from '../managers/GitRefManager.js'
 import { GitShallowManager } from '../managers/GitShallowManager.js'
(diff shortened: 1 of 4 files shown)


src/commands/log.js
```diff
@@ -169,5 +172,53 @@ export async function _log({
...
+    cache,
+    gitdir,
+    trees: [TREE({ ref: commit.oid }), TREE({ ref: parent })],
+    map: async (filepath, [current, previous]) => {
+      const [currentType, previousType] = await Promise.all([
+        current && current.type(),
+        previous && previous.type(),
+      ])
+      if (currentType === 'tree') {
+        if (previousType && previousType !== 'tree') {
```
jcubic, line 205: Question: Do we need to check the 'tree' if Git is not tracking directories, only files?

minutechreview: Git does not track empty directories, but commits do store directory snapshots as tree objects. `_walk` calls `map` for those tree entries before walking their children, so this check prevents a tree OID from being exposed as a changed-file OID. For a blob↔tree replacement, we report only the blob side at that path and still traverse the tree to report its leaf-file changes.

PR #2393, opened 2026-08-07
fix: exclude git system files when expanding a ref

### LINKED ISSUE #767: Cannot create branch with name 'index"
```js
await git.branch({ dir, ref: branch: 'index' });
await git.checkout({ dir, ref: branch: 'index' });

await git.currentBranch({ dir }); // undefined
```

luantaraschi:
Fixes #767.

## I'm fixing a bug

`GitRefManager.resolve` filters the git system files out of its refpath candidates. `expand` does not:

```js
// resolve
const allpaths = refpaths(ref).filter(p => !GIT_FILES.includes(p)) // exclude git system files (#709)

// expand
const allpaths = refpaths(ref)
```

The first candidate `refpaths` produces is the bare name, so `expand({ ref: 'index' })` calls `fs.exists('.git/index')`, finds the index file, and returns `'index'`. Same for `config`, `shallow`, `description` and `commondir`. None of them are refs.

## Why it matters beyond a wrong return value

`checkout` uses `expand` to decide whether the ref names a branch:

```js
const fullRef = await GitRefManager.expand({ fs, gitdir, ref })
if (fullRef.startsWith('refs/heads')) {
  await GitRefManager.writeSymbolicRef({ fs, gitdir, ref: 'HEAD', value: fullRef })
} else {
  // detached head
  await GitRefManager.writeRef({ fs, gitdir, ref: 'HEAD', value: oid })
}
```

`'index'` does not start with `refs/heads`, so checking out a branch called `index` writes a detached HEAD. No error is raised. Commits made afterwards belong to no branch, which is what the reporter saw as `currentBranch()` returning `undefined`.

Reproduced on a fresh repository:

```
HEAD                : 13f4eb0828efbe457d544ee30dfda75eb27e0d75   <- detached
refs/heads/index    : exists
currentBranch       : undefined
expandRef("index")  : index
expandRef("config") : config
```

After the fix:

```
HEAD                : ref: refs/heads/index
currentBranch       : index
expandRef("index")  : refs/heads/index
expandRef("config") : throws NotFoundError
```

## The fix

One line, applying to `expand` the filter `resolve` already has. I left `expandAgainstMap` and `resolveAgainstMap` alone: they look names up in a map of refs, where a repository file can never be a key.

`deleteBranch` is unaffected too, since it normalises to `refs/heads/` before expanding.

## Tests

Two tests, following the checklist in Appendix A of CONTRIBUTING, so each one has a plain and an `-in-submodule` variant:

- `test-GitRefManager.js`: `expand` throws `NotFoundError` for `config`, `index` and `shallow`.
- `test-checkout.js`: checking out a branch named `index` leaves HEAD pointing at `refs/heads/index`.

They are in the first commit and the fix is in the second, so the tests can be watched going from failing to passing.

| | Result |
|---|---|
| First commit only | 4 failed |
| With the fix | 4 passed |

Full suite on Node, before and after:

| | Failures |
|---|---|
| Before the fix | 16 |
| After the fix | 14 |

The difference is exactly these four tests. The 14 that remain fail identically on both sides and are unrelated to refs: five symlink tests, one `autocrlf` test, `clone > should set up the remote tracking branch by default`, and a few others. They are environmental on my machine.

## What I did not verify

I ran the Node suite on Windows 11, not the Chrome or Firefox browser suites. The 14 pre-existing failures listed above are almost certainly the Windows symlink restriction rather than anything about this repository, but I have not confirmed that on Linux, so I am reporting the counts as measured rather than calling the suite green.

I have not run `npm run add-contributor`. Happy to add the entry in this PR if you would rather it landed together.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented Git system files such as `config`, `index`, and `shallow` from being overwritten, deleted, or treated as bare references.
  * Preserved support for valid branches named after system files, including correct checkout and attached `HEAD` behavior.
  * Ensured invalid reference updates are rejected before repository changes or pruning occur.
  * Added regression coverage for standard repositories and submodules.

### DIFF
--- a/__tests__/test-GitRefManager-in-submodule.js
+++ b/__tests__/test-GitRefManager-in-submodule.js
@@ -355,6 +355,110 @@ describe('GitRefManager', () => {
     expect(await fs.read(`${gitdirsmfullpath}/index`)).toEqual(before)
   })
 
+  it('updateRemoteRefs refuses a refspec that targets a git system file', async () => {
+    const { fs, gitdirsmfullpath } =
+      await makeFixtureAsSubmodule('test-checkout')
+    const before = await fs.read(`${gitdirsmfullpath}/index`)
+    // The local side of a refspec is whatever the config says, so a remote
+    // cannot be trusted to translate only into refs/.
+    let error = null
+    try {
+      await GitRefManager.updateRemoteRefs({
+        fs,
+        gitdir: gitdirsmfullpath,
+        remote: 'origin',
+        refs: new Map([
+          ['refs/heads/main', 'e10ebb90d03eaacca84de1af0a59b444232da99e'],
+        ]),
+        symrefs: new Map(),
+        tags: false,
+        refspecs: ['+refs/heads/main:index'],
+      })
+    } catch (err) {
+      error = err
+    }
+    expect(error).not.toBeNull()
+    expect(error.code).toBe(Errors.InvalidRefNameError.code)
+    expect(await fs.read(`${gitdirsmfullpath}/index`)).toEqual(before)
+  })
+
+  it('updateRemoteRefs rejects a bad refspec before pruneTags deletes anything', async () => {
+    const { fs, gitdirsmfullpath } =
+      await makeFixtureAsSubmodule('test-checkout')
+    const oid = 'e10ebb90d03eaacca84de1af0a59b444232da99e'
+    // pruneTags clears refs/tags before the write loop runs, so a refspec that
+    // is refused later takes the tags down with it. The fetch cannot put them
+    // back: it throws before writing anything.
+    let error = null
+    try {
+      await GitRefManager.updateRemoteRefs({
+        fs,
+        gitdir: gitdirsmfullpath,
+        remote: 'origin',
+        refs: new Map([
+          ['refs/heads/main', oid],
+          ['refs/tags/v1.0.0', oid],
+        ]),
+        symrefs: new Map(),
+        tags: true,
+        refspecs: ['+refs/heads/main:index'],
+        pruneTags: true,
+      })
+    } catch (err) {
+      error = err
+    }
+    expect(error).not.toBeNull()
+    expect(error.code).toBe(Errors.InvalidRefNameError.code)
+    expect(
+      await GitRefManager.resolve({
+        fs,
+        gitdir: gitdirsmfullpath,
+        ref: 'refs/tags/v1.0.0',
+      })
+    ).toEqual(oid)
+  })
+
+  it('updateRemoteRefs rejects a bad refspec before prune deletes anything', async () => {
+    const { fs, gitdirsmfullpath } =
+      await makeFixtureAsSubmodule('test-checkout')
+    const oid = 'e10ebb90d03eaacca84de1af0a59b444232da99e'
+    await GitRefManager.writeRef({
+      fs,
+      gitdir: gitdirsmfullpath,
+      ref: 'refs/remotes/origin/stale',
+      value: oid,
+    })
+    // Same hazard on the prune path: a ref the remote no longer advertises is
+    // deleted before the refspec that targets `.git/index` is refused.
+    let error = null
+    try {
+      await GitRefManager.updateRemoteRefs({
+        fs,
+        gitdir: gitdirsmfullpath,
+        remote: 'origin',
+        refs: new Map([['refs/heads/main', oid]]),
+        symrefs: new Map(),
+        tags: false,
+        refspecs: [
+          '+refs/heads/*:refs/remotes/origin/*',
+          '+refs/heads/main:index',
+        ],
+        prune: true,
+      })
+    } catch (err) {
+      error = err
+    }
+    expect(error).not.toBeNull()
+    expect(error.code).toBe(Errors.InvalidRefNameError.code)
+    expect(
+      await GitRefManager.resolve({
+        fs,
+        gitdir: gitdirsmfullpath,
+        ref: 'refs/remotes/origin/stale',
+      })
+    ).toEqual(oid)
+  })
+
   it('expand does not return a git system file', async () => {
     const { fs, gitdirsmfullpath } =
       await makeFixtureAsSubmodule('test-checkout')
--- a/__tests__/test-GitRefManager-in-submodule.js
+++ b/__tests__/test-GitRefManager-in-submodule.js
@@ -308,6 +308,32 @@ describe('GitRefManager', () => {
     await Promise.all(writePromises)
   })
 
+  it('writeRef refuses to write over a git system file', async () => {
+    const { fs, gitdirsmfullpath } =
+      await makeFixtureAsSubmodule('test-checkout')
+    const oid = 'e10ebb90d03eaacca84de1af0a59b444232da99e'
+    const before = await fs.read(`${gitdirsmfullpath}/index`)
+    // `.git/index` is not a ref, and its bare name is a legal one-level ref
+    // name, so an unguarded write lands on the staging area and destroys it.
+    // Canonical git refuses: "cannot lock ref 'index': reference broken".
+    for (const ref of ['config', 'index', 'shallow']) {
+      let error = null
+      try {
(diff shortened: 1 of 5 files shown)


luantaraschi:
Pushed `e352df2a`. I went back over this change looking for what it breaks rather than what it fixes, and found a regression I had introduced. Writing it up in full because the mechanism is not obvious from the diff.

## What I broke

`GitRefManager.exists` calls `expand`. Before this branch, `expand({ ref: 'index' })` found `.git/index` on disk and returned `'index'`, so `exists` reported the ref as taken. `writeRef` checks `exists` and throws `AlreadyExistsError`:

```js
if (!force && (await GitRefManager.exists({ fs, gitdir: updatedGitdir, ref }))) {
  throw new AlreadyExistsError('ref', ref)
}
```

That was accidental protection, and the first commit on this branch removed it. `isValidRef('index', true)` accepts the name, `exists` now returns false, and the write goes through to `fs.write(join(gitdir, 'index'), value)`.

Measured on a fresh repository, `writeRef({ ref: 'index', value: oid })`:

| | result | `.git/index` |
|---|---|---|
| `main` | throws `AlreadyExistsError` | 104 bytes, `DIRC`, intact |
| this branch before `e352df2a` | resolves, no error | **41 bytes, staging area destroyed** |

The hazard is not new. `force: true` skipped the `exists` check on `main` too, so the same write was already reachable. What changed is that it stopped needing `force`.

## The guard

`assertWritableRef` in `GitRefManager`, called by `writeRef` and `writeSymbolicRef`, refuses when the ref is exactly one of `GIT_FILES`. It covers the `force` path as well, so this closes the older hazard too.

I checked this against real git before picking the behaviour:

```
$ git update-ref index <sha>
fatal: update_ref failed for ref 'index': cannot lock ref 'index': unable to resolve reference 'index': reference broken
```

`.git/index` is left untouched. Refusing matches canonical git rather than being something I invented.

I put it in the manager rather than in `api/writeRef.js` so every path is covered. I checked all fifteen internal call sites: they pass either `HEAD` or a `refs/...` name, none of which is in `GIT_FILES`, so nothing internal changes.

## Verification

Two tests added, plain and `-in-submodule`, asserting the throw and that `.git/index` is byte-identical afterwards. Both fail on the commit before the guard.

Full suite: 14 failures before and after, the same set every time, all of them the Windows symlink restriction and the mock-server clone test. Test count goes 1295 to 1299.

I also confirmed the legitimate case still works end to end: `branch({ ref: 'index' })` then `checkout({ ref: 'index' })` leaves `currentBranch()` returning `index`, which is the bug this PR set out to fix.

## Two things I found and deliberately left out

**`GIT_FILES` is a blacklist and it is incomplete.** `fs.exists` returns true for directories, and `init` creates `.git/hooks`, `.git/info`, `.git/objects`, `.git/refs`. A branch named `hooks` or `objects` reproduces the identical detached-HEAD bug even with this PR applied. Same for `logs`, `worktrees`, `modules`, `packed-refs`, `COMMIT_EDITMSG`, `FETCH_HEAD`, `ORIG_HEAD`.

Canonical git does not blacklist. Per gitrevisions it accepts the bare candidate only for `refs/...` paths and an all-caps pseudo-ref set. Switching `refpaths` to that whitelist would close the whole class. I checked feasibility: the only bare non-`refs/` literal passed to `resolve` or `expand` anywhere in `src` is `'HEAD'`. So it is viable, but it is a rewrite of ref resolution rather than a bug fix, and I would rather it be its own PR with its own review. Say the word if you would prefer it here instead.

**`expandAgainstMap` and `resolveAgainstMap` also lack the filter.** I left them alone on purpose. Their callers in `push.js` and `fetch.js` pass the remote ref advertisement, whose keys are always full refs, so a bare `index` key cannot occur. Filtering there would be dead code.

luantaraschi:
Both valid. Fixed in `0c86bf71`, and I checked each one against the running code rather than taking them as read, because the second one contradicted an assumption I had made in my own earlier comment.

**`deleteRefs`.** Confirmed, and it is worse than the `writeRef` path: that one overwrote `.git/index`, this one removes it.

```
.git/index before: 104 bytes
deleteRef({ ref: 'index' }) resolved
.git/index after:  DELETED
```

Worth noting for whoever reviews this: it reproduces identically on `main`, so it is not something this branch introduced. `deleteRefs` calls `fs.rm(join(gitdir, ref))` with no check at all and always has.

**`updateRemoteRefs`.** I had argued in the comment above that the keys are always refspec translations and therefore always start with `refs/`, so a guard there would be dead code. That was wrong, and reading `GitRefSpec.translate` is what showed it:

```js
translate(remoteBranch) {
  if (this.matchPrefix) {
    if (remoteBranch.startsWith(this.remotePath)) {
      return this.localPath + remoteBranch.replace(this.remotePath, '')
    }
  } else {
    if (remoteBranch === this.remotePath) return this.localPath
  }
```

`localPath` is taken verbatim from `remote.<name>.fetch`, so a config line of `+refs/heads/main:index` translates to exactly the key that hurts. Not structurally impossible, just config-dependent.

Both guards go through the same `assertWritableRef`. I put the `updateRemoteRefs` check in its own pass over the keys before the write loop, rather than inline, so a bad refspec fails before anything is written instead of halfway through.

Three tests now on this branch, each with its `-in-submodule` twin: `writeRef`, `deleteRefs`, and `updateRemoteRefs` driven through an explicit `refspecs: ['+refs/heads/main:index']`. Each asserts the throw and that `.git/index` is byte-identical afterwards, and each fails without its guard.

Full suite: 14 failures before and after, the same set every run, all of them the Windows symlink restriction and the mock-server clone test. Test count 1295 to 1302.

jcubic:
Can you fix the last Coderabbit suggestion?

luantaraschi:
Done, pushed `24473f4b` and `1b6eb8c2` on top of your `42f84608`.

I checked the suggestion against the code before applying it, and as written it is incomplete. It says to move the validation loop to immediately after `actualRefsToWrite` is built. That closes `prune`, but `pruneTags` runs earlier still, before the map has a single entry in it, so the tags are already gone by the time the loop could run.

Measured on a fresh `test-checkout` fixture, `updateRemoteRefs` with `pruneTags: true` and a refspec of `+refs/heads/main:index`:

```
before: refs/tags/v1.0.0 -> e10ebb9
throws InvalidRefNameError
after:  NotFoundError: Could not find refs/tags/v1.0.0
```

The fetch cannot put it back either, because the throw happens before the write loop.

## What I did instead

Translating a refspec is pure, so the destinations get worked out at the top of the function and checked there, ahead of every deletion. The loops that populate `actualRefsToWrite` stay where they were and consume the precomputed values, so the map is built in the same order as before and `prune` still sees the same set.

The only thing added between the check and the writes is the `if (tags)` block, and its keys are always `refs/tags/...`, which cannot be a system file. Nothing gets past the check by being added late.

I moved the symref handling up together with the translation rather than the `translate()` call on its own, so the `symtarget` guard travels with it. Validating the raw `translate()` output would have rejected a symref whose target does not translate, where today nothing is written for it at all. That would have been a behaviour change riding along with the fix.

## Verification

Four tests, two cases with their `-in-submodule` twins: `pruneTags` and `prune`, each asserting the throw and that the pre-existing ref still resolves afterwards. All four fail on `0c86bf71`, and they fail on the assertion about the ref, not on the error code.

Full suite: 14 failures before and after, the same set every run, all of them the Windows symlink restriction and the clone test that depends on it. Test count 1302 to 1309.

## Two notes on the test files

The `-in-submodule` twin of `updateRemoteRefs refuses a refspec that targets a git system file` was never there. I said in an earlier comment that all three guards had twins and that was wrong, only `writeRef` and `deleteRefs` did. It is there now.

I also mirrored your new test into the submodule suite, since every other case in that file has a twin. The internal `GitRefManager` calls need `gitdirsmfullpath` there, because only the public API resolves the submodule's `.git` file. I checked both copies have teeth by changing the filter to drop candidates by their last path segment instead of the bare name, which is the plausible wrong version of it, and both fail.

PR #2394, opened 2026-08-07
fix: apply .gitignore only to files that are not tracked

### LINKED ISSUE #765: git.add a file which is tracked, but it is matched the .gitignore files
Please be sure to mention:
- Node
- 0.55.4

luantaraschi:
Fixes #765.

## I'm fixing a bug

`.gitignore` is a rule about untracked files. Once a path is in the index, canonical git keeps staging changes to it, keeps reporting its real state, and stops matching it with `check-ignore`:

```
$ git init -q . && echo x > f.txt && git add f.txt && git commit -qm i
$ echo "f.txt" > .gitignore && echo y >> f.txt

$ git status --short
 M f.txt

$ git add f.txt && git status --short
M  f.txt

$ git check-ignore -v f.txt
(no match)
```

isomorphic-git checks the ignore rules before it consults the index, in two places:

- `api/add.js` returns early, so the change is never staged and no error is raised.
- `api/status.js` runs `isIgnored` as its first statement, before it looks at HEAD or the index, and returns `'ignored'` for a tracked file.

`api/statusMatrix.js` already gets this right. Its guard is `if (!head && !stage && workdir)`, under the comment "Ignore ignored files, but only if they are not already tracked". And `api/add.js` carried the rule as a TODO:

```js
// TODO: Should ignore UNLESS it's already in the index.
```

So this is one rule the project has already decided on, implemented at one of the three call sites that need it.

## The fix

`add` consults the index it is already holding:

```diff
-  // TODO: Should ignore UNLESS it's already in the index.
   filepath = Array.isArray(filepath) ? filepath : [filepath]
   const promises = filepath.map(async currentFilepath => {
-    if (!force) {
+    if (!force && !index.has({ filepath: currentFilepath })) {
```

`status` moves its ignore check below the HEAD and index lookups and guards it the way `statusMatrix` does. Untracked ignored files return `'ignored'` exactly as before.

I did not touch `statusMatrix`, and I did not add a workdir condition to `status`. `statusMatrix` needs one because it walks the working directory; `status` is called with an explicit filepath, and adding it would change what an absent untracked ignored path returns, which is not what this issue is about.

## Tests

Four tests, each with the `-in-submodule` variant that Appendix A of CONTRIBUTING asks for:

- `test-add.js`: track `i.txt`, then ignore it, then change it on disk. `add` must stage the new content. The test reads the staged blob rather than asking `status`, so it does not depend on the second fix.
- `test-status.js`: same setup, `status` must return `*added` rather than `ignored`.

Tests are in the first commit and the fix is in the second, so they can be watched going from failing to passing.

| | Result |
|---|---|
| First commit only | 4 failed |
| With the fix | 44 passed across the 4 suites |

The control failures are the damage itself, not a proxy for it:

```
● add › ignored file that is already tracked
  Expected: "changed after being ignored"
  Received: "ignore me!"

● status › tracked file that matches a .gitignore rule
  Expected: "*added"
  Received: "ignored"
```

The staged blob still holds the pre-ignore content, which is the silent part: `add` resolves, no error is raised, and the change is gone.

Full suite on Node, before and after:

| | Failures |
|---|---|
| Before the fix | 18 |
| After the fix | 14 |

The difference is exactly these four. The 14 that remain fail identically on both sides and have nothing to do with ignore rules: five symlink tests, one `autocrlf` test, `clone > should set up the remote tracking branch by default`, and a few others.

## What I did not verify

I ran the Node suite on Windows 11, not the Chrome or Firefox suites. The 14 pre-existing failures are very likely the Windows symlink restriction, but I have not confirmed that on Linux, so I am reporting the counts as measured rather than calling the suite green.

I have not run `npm run add-contributor`. Happy to add the entry here if you would rather it landed in the same PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tracked files that later match a `.gitignore` rule can now be staged when modified.
  * Status results now correctly report changes to tracked files, even when they are subsequently ignored.
* **Tests**
  * Added coverage for staging and status behavior in repositories and submodules.

### DIFF
--- a/__tests__/test-add-in-submodule.js
+++ b/__tests__/test-add-in-submodule.js
@@ -275,4 +275,31 @@ describe('add', () => {
       'added'
     )
   })
+
+  it('ignored file that is already tracked', async () => {
+    // Setup
+    const { fs, dir, gitdir } = await makeFixtureAsSubmodule('test-add')
+    await init({ fs, dir, gitdir })
+    // Track the file before it becomes ignored. That is how a file ends up
+    // tracked and ignored at the same time in a real repository.
+    await add({ fs, dir, gitdir, filepath: 'i.txt' })
+    await writeGitIgnore(fs, dir)
+    await fs.write(dir + '/i.txt', 'changed after being ignored')
+    // Test
+    await add({ fs, dir, gitdir, filepath: 'i.txt' })
+    // .gitignore applies to untracked files only, so canonical git stages this
+    // change instead of silently doing nothing.
+    const staged = await walk({
+      fs,
+      dir,
+      gitdir,
+      trees: [STAGE()],
+      map: async (filepath, [stage]) =>
+        filepath === 'i.txt' && stage ? stage.oid() : undefined,
+    })
+    const { blob } = await readBlob({ fs, dir, gitdir, oid: staged[0] })
+    expect(Buffer.from(blob).toString('utf8')).toEqual(
+      'changed after being ignored'
+    )
+  })
 })
--- a/__tests__/test-add-in-submodule.js
+++ b/__tests__/test-add-in-submodule.js
@@ -174,6 +174,35 @@ describe('add', () => {
     await add({ fs, dir, filepath: 'i.txt', force: true })
     expect((await listFiles({ fs, dir })).length).toEqual(1)
   })
+  it('tracked file inside an ignored folder', async () => {
+    // Setup
+    const { fs, dir, gitdir } = await makeFixtureAsSubmodule('test-add')
+    await init({ fs, dir, gitdir })
+    // Track one file inside the folder before the folder becomes ignored.
+    await add({ fs, dir, gitdir, filepath: 'js_modules/awesome/index.js' })
+    await writeGitIgnore(fs, dir)
+    await fs.write(dir + '/js_modules/awesome/index.js', 'changed')
+    await fs.write(dir + '/js_modules/untracked.js', 'never staged')
+    // Test
+    await add({ fs, dir, gitdir, filepath: '.' })
+    // Canonical git recurses into an ignored folder for the sake of the paths
+    // it already tracks, and stages those while leaving the rest alone.
+    const staged = await walk({
+      fs,
+      dir,
+      gitdir,
+      trees: [STAGE()],
+      map: async (filepath, [stage]) =>
+        filepath === 'js_modules/awesome/index.js' && stage
+          ? stage.oid()
+          : undefined,
+    })
+    const { blob } = await readBlob({ fs, dir, gitdir, oid: staged[0] })
+    expect(Buffer.from(blob).toString('utf8')).toEqual('changed')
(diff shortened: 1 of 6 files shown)


luantaraschi:
Pushed `891a0c6e`. I went back over this against real git rather than against my own reading of it, and the first pass was incomplete.

## What I missed

I only compared the exact path against the index. That fixes a tracked file, and leaves a tracked file inside an ignored **folder** broken. `addToIndex` reaches the folder, finds no index entry under that exact name, sees it is ignored, and returns before it can recurse.

That matters because it breaks `add({ filepath: '.' })`, which is the most common call there is.

Real git, same repository state:

```
$ echo js_modules > .gitignore
$ git add .
$ git status --short
A  .gitignore
M  js_modules/tracked.txt

$ git show :js_modules/tracked.txt
v2 changed
```

isomorphic-git with the first commit only:

```
staged content of js_modules/tracked.txt: "v1"
```

The old content, silently.

## The fix

`isTracked` now answers true for a folder that holds a tracked path, so the walk enters it. The recursion re-checks every child, and `GitIgnoreManager.isIgnored` returns true for a child of an ignored folder on its own, so untracked siblings inside the folder are still skipped. I verified the mixed case: `js_modules/awesome/index.js` gets staged, `js_modules/untracked.js` does not, which is what real git does.

On cost: the exact lookup is a `Map` hit and runs first, so a tracked file answers immediately. The prefix scan over `entriesMap` only runs for a path that is not itself tracked, which for a normal `add .` means folders. `GitIndex.delete` already uses the same `startsWith(filepath + '/')` idiom, so it is not a new pattern here.

## A behaviour change I should have declared the first time

A file that is tracked, ignored, and **deleted from the working tree** now throws `NotFoundError` where it used to resolve silently. Measured three ways:

| | result |
|---|---|
| `main` | resolves, does nothing |
| this branch | throws `NotFoundError` |
| canonical git | stages the deletion, `D i.txt` |

Neither the old nor the new behaviour matches git, which stages the removal. The new one is at least consistent with how `add` treats every other tracked file that is missing from disk, since `src/api/add.js` throws `NotFoundError` for any of them. Staging deletions is a separate feature and I did not add it here. It is only reachable through an explicit `filepath`, never through `add({ filepath: '.' })`, which discovers paths by `readdir`.

## Verification

Six tests now, three cases with their `-in-submodule` twins: the tracked file, the tracked file inside an ignored folder, and `status` on a tracked ignored path. All fail on the test commit and pass with the fixes.

Full suite: 14 failures before and after, the same set both times, all of them the Windows symlink restriction and the mock-server clone test. Test count 1295 to 1299.

## Still out of scope, and worth flagging

`src/utils/walkerToTreeEntryMap.js` runs the same unguarded ignore check twice, at `writeTreeChanges` and `applyTreeChanges`, which back `stash` and `cherryPick`. So `stash` will also drop a tracked-but-ignored file. I did not touch it: it is a different call path with different tests, and this diff is already two commits wide. Happy to send it separately.

PR #2395, opened 2026-08-07
fix: honour core.autocrlf when status hashes the working copy

### LINKED ISSUE #1275: Files marked as *modified after clone
I am using Node, on Windows

## Problem

On Windows and with some repositories, isogit is telling me that all files are modified just after cloning the repo.

This is blocking me because I can't find which files have been changed locally since the last commit

## Steps to reproduce

```
npm i -g isomorphic-git
git clone https://github.com/acailly/zDemocracy-lowtech-data.git
cd zDemocracy-lowtech-data
isogit status --filepath="README.md"
> "*modified"
```

This is not reproduced on other repo like https://github.com/isomorphic-git/isomorphic-git.git

I do not reproduce it on the Ubuntu executed in WSL2 (Windows Subsystem for Linux)

luantaraschi:
Fixes #1275.

## I'm fixing a bug

When the working tree holds CRLF and `core.autocrlf=true`, `status` reports `*modified` for a file nobody touched, disagreeing with both `statusMatrix()` and `git status`. The blob is stored with LF, `status` reads the file with no normalisation, hashes the CRLF bytes, and gets an oid that can never match the one in the index.

One correction to my first version of this description, which said a checkout leaves CRLF on disk. That is canonical git, not this library. `autocrlf` appears once in `FileSystem`, inside `read`, and `checkout` calls `fs.write(filepath, object)` with no options, so isomorphic-git never writes CRLF anywhere. A clone performed by this library cannot produce the reported state on its own; it comes from canonical git, an editor, or a syncing tool. The defect in `status` is the same either way.

`GitWalkerFs`, which backs `walk` and `statusMatrix`, already reads the right way:

```js
const config = await this._getGitConfig(fs, gitdir)
const autocrlf = await config.get('core.autocrlf')
content = await fs.read(`${dir}/${entry._fullpath}`, { autocrlf })
```

`api/status.js` did not:

```js
const object = await fs.read(join(dir, filepath))
```

So the two disagree about the same file. On Windows 11, same repository, same moment:

| | result |
|---|---|
| `statusMatrix()` | `[['a.txt', 1, 1, 1]]`, unmodified |
| `status({ filepath: 'a.txt' })` | `*modified` |

@acailly diagnosed this down to the bytes back in 2020, comparing the two hashes and finding `0a` on one side and `0d0a` on the other. The `autocrlf` support that landed since covers `add`, `walk` and `statusMatrix`, and `status` was left out.

## The fix

`status` reads through the same normalisation:

```diff
-        const object = await fs.read(join(dir, filepath))
+        const config = await GitConfigManager.get({ fs, gitdir: updatedGitdir })
+        const autocrlf = await config.get('core.autocrlf')
+        const object = await fs.read(join(dir, filepath), { autocrlf })
```

It sits inside the branch that already decided to hash the file, so the fast path where `compareStats` matches is untouched and nothing extra is read when the stats are clean.

## Tests

One test with the `-in-submodule` variant Appendix A of CONTRIBUTING asks for. Commit a file with LF under `core.autocrlf=true`, then write CRLF to the working copy explicitly, since nothing in this library would put it there.

**The assertion order is load-bearing, and my first version of this test was wrong.** It asserted `statusMatrix` first, and it passed without the fix. `statusMatrix` refreshes the index stat cache, and once the cached stats match the CRLF file on disk, `status` short-circuits on `compareStats` and returns the indexed oid without hashing anything. The bug was hidden by the line above it. `status` is checked first now, and there is a comment in the test saying why, so nobody reorders it later.

| | Result |
|---|---|
| First commit only | 2 failed |
| With the fix | 8 passed across both suites |

Control failure:

```
● status › honours core.autocrlf when hashing the working copy
  Expected: "unmodified"
  Received: "*modified"
```

Full suite on Node, Windows 11:

| | Failures |
|---|---|
| Before the fix | 16 |
| After the fix | 14 |

The 14 that remain fail identically on both sides. Thirteen of them are one thing: a fixture symlink comes back as `mode 33188` instead of `40960`, because creating symlinks on Windows needs elevation. The fourteenth is `clone > should set up the remote tracking branch by default`, which fails with a `MultipleGitError` against the local mock server.

One note on flakiness, since I saw it and would rather say so: in one of four full runs, `commit > create signed commit` failed on a signature check. It passed on the other three runs and passes on its own, and it has nothing to do with this change, but it is not deterministic on my machine.

## Overlap with #2394

That PR also touches `api/status.js` and `__tests__/test-status.js`. I test-merged the two branches: `src/api/status.js` merges cleanly, because the changes are in different parts of the function. `__tests__/test-status.js` conflicts, only because both branches append an `it()` at the end of the same `describe`. Whichever one you take first, I will rebase the other and repost the control numbers.

## What I did not verify

I ran the Node suite on Windows 11, which is the platform this bug is about, and not the Chrome or Firefox suites. `core.autocrlf=input` is still ignored after this change, deliberately. `FileSystem.js` gates on `options.autocrlf === 'true'`, so `input`, `yes`, `on` and `True` all fall through. For hashing purposes `input` should behave like `true`, but widening that gate changes `add` and the walker too, which is a behaviour change rather than a bug fix and deserves its own review. Naming it so this PR is not read as closing out autocrlf entirely.

Two sibling call sites read the working copy unnormalised the same way: `src/api/resetIndex.js` and `src/api/updateIndex.js`. The second is worse than the bug reported here, because the unnormalised bytes go on to `_writeObject`, so a CRLF blob enters the object database. I left both out to keep the diff reviewable and would rather raise `updateIndex` as its own issue.

I have not run `npm run add-contributor`. Happy to add the entry here if you would rather it landed in the same PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed status checks incorrectly reporting files as modified when CRLF line endings are used with `core.autocrlf=true`.
  * Updated status comparisons to normalize working-tree line endings consistently.
  * Ensured consistent behavior for both regular repositories and submodules.

### DIFF
--- a/__tests__/test-status-in-submodule.js
+++ b/__tests__/test-status-in-submodule.js
@@ -139,8 +139,9 @@ describe('status', () => {
       message: 'initial',
       author: { name: 'Test', email: 'test@example.com' },
     })
-    // What a checkout under core.autocrlf=true leaves on a Windows disk: the
-    // blob is stored with LF and the working copy carries CRLF.
+    // The blob is stored with LF and the working copy carries CRLF. This
+    // library never writes CRLF, so on a real machine the working copy got
+    // that way from canonical git, an editor, or a syncing tool.
     await fs.write(path.join(dir, 'a.txt'), 'one\r\ntwo\r\n')
     // Test
     // status has to come first. statusMatrix refreshes the index stat cache,
--- a/__tests__/test-status-in-submodule.js
+++ b/__tests__/test-status-in-submodule.js
@@ -1,7 +1,15 @@
 /* eslint-env node, browser, jasmine */
 import * as path from 'path'
 
-import { status, add, init, remove } from 'isomorphic-git'
+import {
+  status,
+  add,
+  commit,
+  init,
+  remove,
+  setConfig,
+  statusMatrix,
+} from 'isomorphic-git'
 
 import { makeFixtureAsSubmodule } from './__helpers__/FixtureFSSubmodule.js'
 
@@ -115,5 +123,36 @@ describe('status', () => {
     expect(await status({ fs, dir, gitdir, filepath: 'i.txt' })).toEqual(
       '*added'
     )
+  })
+
+  it('honours core.autocrlf when hashing the working copy', async () => {
+    // Setup
+    const { fs, dir, gitdir } = await makeFixtureAsSubmodule('test-empty')
+    await init({ fs, dir, gitdir })
+    await setConfig({ fs, dir, gitdir, path: 'core.autocrlf', value: true })
+    await fs.write(path.join(dir, 'a.txt'), 'one\ntwo\n')
+    await add({ fs, dir, gitdir, filepath: 'a.txt' })
+    await commit({
+      fs,
+      dir,
+      gitdir,
+      message: 'initial',
+      author: { name: 'Test', email: 'test@example.com' },
+    })
+    // What a checkout under core.autocrlf=true leaves on a Windows disk: the
+    // blob is stored with LF and the working copy carries CRLF.
+    await fs.write(path.join(dir, 'a.txt'), 'one\r\ntwo\r\n')
+    // Test
+    // status has to come first. statusMatrix refreshes the index stat cache,
+    // and once the cached stats match the CRLF file on disk, status returns the
+    // indexed oid without hashing anything, which hides the bug.
+    expect(await status({ fs, dir, gitdir, filepath: 'a.txt' })).toEqual(
+      'unmodified'
+    )
+    // GitWalkerFs already reads with the autocrlf option, so statusMatrix
+    // agrees the file is untouched either way.
+    expect(await statusMatrix({ fs, dir, gitdir })).toEqual([
+      ['a.txt', 1, 1, 1],
+    ])
   })
 })
--- a/__tests__/test-status.js
+++ b/__tests__/test-status.js
@@ -1,7 +1,15 @@
 /* eslint-env node, browser, jasmine */
 import * as path from 'path'
 
-import { status, add, init, remove } from 'isomorphic-git'
+import {
+  status,
+  add,
+  commit,
+  init,
+  remove,
+  setConfig,
+  statusMatrix,
+} from 'isomorphic-git'
 
 import { makeFixture } from './__helpers__/FixtureFS.js'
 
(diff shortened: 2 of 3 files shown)


luantaraschi:
Pushed `53cf1da2` and rewrote the description above. The code is unchanged; what was wrong was the framing.

I had written that a checkout under `core.autocrlf=true` leaves CRLF on the disk. That is canonical git, not this library. `autocrlf` appears exactly once in `src/models/FileSystem.js`, inside `read`, and `checkout` calls `fs.write(filepath, object)` with no options, so nothing in isomorphic-git ever writes CRLF. A clone performed by this library cannot produce the reported state on its own.

That matters for how the issue reads, so I would rather correct it than let it stand: the working copy got its CRLF from canonical git, an editor, or a syncing tool. The defect in `status` is identical either way, and the test writes the CRLF explicitly rather than relying on a checkout to produce it.

Two related things I added to the description while I was at it, both deliberately out of this diff:

`core.autocrlf=input` stays ignored. The gate is `options.autocrlf === 'true'`, so `input`, `yes`, `on` and `True` all fall through. For hashing, `input` should behave like `true`, but widening that gate changes `add` and the walker too. That is a behaviour change, not a bug fix.

`src/api/resetIndex.js` and `src/api/updateIndex.js` read the working copy unnormalised the same way. `updateIndex` is the worse of the two, because those bytes go on to `_writeObject`, so a CRLF blob lands in the object database. That one is arguably more serious than what this PR fixes, and I would rather raise it as its own issue than bury it here. Say the word if you would prefer it bundled.

jcubic:
Please resolve the conflicts.

luantaraschi:
Rebased onto `main`. The conflicts came from #2394, which landed in the same two status test files and reworked `src/api/status.js` around them.

Both conflicts were additive rather than competing. #2394 added `tracked file that matches a .gitignore rule` and this branch adds `honours core.autocrlf when hashing the working copy`, in the same spot at the end of the describe, and each needed different imports. I kept both tests and took the union of the imports.

`src/api/status.js` applied without a conflict, but #2394 restructured that function, so I checked the result rather than trusting the clean apply. The ignore handling #2394 added runs earlier, before the stat, while this change sits inside `getWorkdirOid` in the branch that actually hashes the file. They do not overlap.

Verification, on the rebased branch:

```
── src/api/status.js reverted to main ──
  ● status › honours core.autocrlf when hashing the working copy
  ● status › honours core.autocrlf when hashing the working copy
Tests: 2 failed, 8 passed, 10 total

── restored ──
Tests: 10 passed, 10 total
```

The two failures are this branch's tests, one per file. The other eight pass either way, which is what I wanted to see: #2394's gitignore test still passes on the merged result, so the resolution did not quietly drop its behaviour.

I made one mistake worth mentioning, in case it matters to how you read the numbers above. My first attempt to verify used `git stash` on `src/api/status.js`, which after a rebase stashes nothing because the change is committed rather than in the working tree. That run reported the tests passing without the fix, which was wrong. The output above is from reverting the file to `main` instead.

Full suite: 14 failures before and after, the same set, all of them the Windows symlink restriction and the clone test that depends on it.

luantaraschi:
﻿The six browser jobs on this PR are red, and they are red on `main` too. The merge commit for #2394 (`f8bf9dba`) shows the same set failing: `ChromeHeadlessNoSandbox`, `FirefoxHeadless`, `sl_edge`, `sl_safari`, `sl_ios_safari`, `bs_android_chrome`. Node builds pass on both.

So this is not fallout from the rebase. Flagging it so the red does not read as this branch's doing, and in case the browser lane going down is news to you.

jcubic:
The GitHub workflow matrix is not working yet; I'm in the process of making them work. I only wanted to see Windows tests.

PR #2408, opened 2026-08-12
docs(readme): fix ES import example using unexported 'plugins'

### LINKED ISSUE #1526: Examples suggest importing 'plugins', which is not exported
The frontpage of the Git repo references importing 'plugins'. But plugins is no longer used by isogit as far as I can tell.

CleanShot 2022-03-31 at 15 05 26@2x

But plugins isn't exported, so it naturally can't be imported, and so it fails to run for me.

I imagine maybe a version got rid of plugins at some point? What's the alternative? Is there a canonical React integration example I should be referring to?

----

Old issue text, ignore:

Trying to get this running in my React project. I'm working from [a React/isogit example (v0.5x)](https://stackblitz.com/edit/isomorphic-git?file=GitService.tsx) which imports 'plugins'; it's not the example I'm using, but the same is done in an example on [this page here on the repo (see screenshot below)](https://github.com/isomorphic-git/isomorphic-git) (plugins is also used in [this v0.74 example](https://isomorphic-git.org/docs/en/0.74.0/plugin_fs), which if I'm reading it right is largely doing the same thing).

Vansh1811:
Closes #1526. The named-import example in the Getting Started section imported `plugins`, which isomorphic-git no longer exports, causing the example to fail with an import error. Replaced it with `add`, a function that is actually exported, so the example works as written.

## I'm fixing a bug or typo

- [x] this is not my first time contributing, so no README change needed
- [x] this is a docs-only change (README.md), squash merge is fine

## Summary by CodeRabbit

* **Documentation**
  * Updated the ES module usage example to demonstrate importing `add` from `isomorphic-git`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ import git from 'isomorphic-git'
 // or
 import * as git from 'isomorphic-git'
 // or
-import {plugins, clone, commit, push} from 'isomorphic-git'
+import {add, clone, commit, push} from 'isomorphic-git'
 ```
 
 View the full [Getting Started guide](https://isomorphic-git.github.io/docs/quickstart.html) on the docs website.


luantaraschi (approved):
Checked this against the current build. `add`, `clone`, `commit`, and `push` are all exported, while `plugins` was removed with the v1 API changes tracked in #1526. The example is valid with this change, and all checks are passing.

PR #2383, opened 2026-07-21
restore karma matrix tests

jcubic:
## I'm fixing a bug or typo

- [ ] if this is your first time contributing, run `npm run add-contributor` and follow the prompts to add yourself to the README
- [ ] squash merge the PR with commit message "fix: [Description of fix]"

## I'm adding a parameter to an existing command X:

- [ ] add parameter to the function in `src/api/X.js` (and `src/commands/X.js` if necessary)
- [ ] document the parameter in the JSDoc comment above the function
- [ ] add a test case in `__tests__/test-X.js` if possible
- [ ] if this is your first time contributing, run `npm run add-contributor` and follow the prompts to add yourself to the README
- [ ] read "Appendix A submodules" in the CONTRIBUTING.md document. Be sure to include submodule tests.

## I'm adding a new command:

- [ ] add as a new file in `src/api` (and `src/commands` if necessary)
- [ ] add command to `src/index.js`
- [ ] update `__tests__/test-exports.js`
- [ ] create a test in `src/__tests__`
- [ ] document the command with a JSDoc comment
- [ ] add page to the Docs Sidebar `website/sidebars.json`
- [ ] add page to the v1 Docs Sidebar `website/versioned_sidebars/version-1.x-sidebars.json`
- [ ] if this is your first time contributing, run `npm run add-contributor` and follow the prompts to add yourself to the README
- [ ] squash merge the PR with commit message "feat: Added 'X' command"
- [ ] read "Appendix A submodules" in the CONTRIBUTING.md document. Be sure to include submodule tests.

## Summary by CodeRabbit

* **Tests**
  * Switched browser testing to Karma for wider multi-browser coverage (local and cloud).
  * Added Karma reporting to summarize failures, post PR comments, and persist successful browser lists for faster reruns.
  * Improved fixture isolation by resetting shared ZenFS overlays between tests.
  * Added server-only Jasmine coverage for `discoverGitdir` and `version()` parity.
* **Bug Fixes**
  * Improved fixture base/URL handling to ensure consistent ZenFS fetch behavior.
* **Chores**
  * Removed Jest/Puppeteer browser test wiring and updated scripts/config to use Karma instead.

PR #2411, opened 2026-08-13
fix: avoid Windows path alias mismatch in symlink map

luantaraschi:
## Summary

Run `git ls-files` from the fixture root so Git returns paths relative to the same directory used by Node.

## Why

The regression test added in #2410 fails on the GitHub Windows runner. Git reports the temporary profile through its short `RUNNER~1` alias, while Node keeps the `runneradmin` path. Calling `path.relative()` across those two forms produces a pathspec outside the repository, which `git ls-files` rejects.

Running the command from the fixture root avoids that path conversion and keeps the map entries relative to `root` on every platform.

## Validation

- `server-only.test-make-symlinks-map.js`: 1 passed
- `nps build.indexjson`: preserved the symlink map and removed its fetch-index entry
- ESLint passed for the helper and regression test
- `git diff --check` passed
- The full Node suite passed 185 of 191 suites and 1,313 of 1,345 tests locally. The remaining 14 failures require Windows symlink privileges and reproduced unchanged on the clean `karma-matrix` head.

Follow-up to #2410. This fixes the Windows failure currently reported on #2383.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved symlink discovery for fixture-based workflows.
  * Ensured symlink paths are resolved correctly relative to the fixture root.

### DIFF
--- a/__tests__/__helpers__/make-symlinks-map.cjs
+++ b/__tests__/__helpers__/make-symlinks-map.cjs
@@ -49,22 +49,17 @@ function makeSymlinksMap(root) {
 
   // With core.symlinks=false, Git writes a symlink's target to a regular file.
   // Read the index to preserve those links in the browser fixture map.
-  const repoRoot = execFileSync('git', ['rev-parse', '--show-toplevel'], {
-    cwd: root,
-    encoding: 'utf8',
-  }).trim()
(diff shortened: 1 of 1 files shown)


jcubic:
@coderabbitai review

